PR #2, opened 2020-11-03
add option to run `cargo clippy` instead of `cargo check`

dicej:


Canop:
Good one. I had thought about the tests (but decided against because of the time they often take) but hadn't thought about clippy.

PR #14, opened 2020-11-04
[WIP] Initial support for workspaces

nikhilmitrax:
Adds support for workspaces by parsing `cargo metadata` and filtering the targets to find the `src/` directories of the workspace members.

Adds dependancy on `serde` and `serde_json` to parse the metadata.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -20,7 +20,7 @@ fn find_folders_to_watch(cargo_filepath: &PathBuf) -> Result<Vec<PathBuf>> {
         .manifest_path(cargo_filepath)
         .exec()?;
 
-    let mut folders_to_watch = vec![];
+    let mut paths_to_watch = vec![];
 
     for item in metadata.packages {
         if item.source.is_none() {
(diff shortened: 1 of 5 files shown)


jyn514:
For anything complicated like this I highly recommend https://docs.rs/cargo_metadata. It has the format predefined which avoids typos or omissions, and also moves the slow serde derives into a separate crate.

nikhilmitrax:
Changed. It now directly relies on `cargo_metadata`

src/app.rs
```diff
@@ -10,22 +10,44 @@ use {
...
 
+
+fn find_folders_to_watch(cargo_filepath: &PathBuf) -> Result<Vec<PathBuf>> {
+    let metadata = MetadataCommand::new().manifest_path(cargo_filepath).exec()?;
+
+    let mut folders_to_watch = vec![];
+
+    for item in metadata.packages {
+        if item.source.is_none() {
+            // We're only concerned with items where the source is None
```
jyn514, line 26:
```suggestion
            // We're only concerned with items that are part of the current workspace.
```

src/app.rs
```diff
@@ -10,22 +10,44 @@ use {
...
+            for target in item.targets {
+                if target.src_path.ends_with("lib.rs") || target.src_path.ends_with("main.rs") {
```
jyn514, line 28: This doesn't seem like the right criteria. You can explicitly rename both of these to be something else. Instead you should look at `crate_type` and `kind`; unfortunately the logic is a little hairy: https://github.com/deadlinks/cargo-deadlinks/blob/master/src/main.rs#L157

Your logic will be slightly different because you don't care about documentation - I'm not sure exactly what this check is for tbh.

src/app.rs
```diff
@@ -10,22 +10,44 @@ use {
...
+                    // targets can contain the build script as well, so we need to eliminate them.
+
+                    let target_folder = target.src_path.parent().expect("parent of a target folder is a root folder");
+                    if target_folder.ends_with("src") {
```
jyn514, line 32: This also doesn't seem right, `src` is just a convention and not required.

src/app.rs
```diff
@@ -110,3 +135,22 @@ pub fn run(w: &mut W, args: Args) -> Result<()> {
     }
     Ok(())
 }
+
+fn load_from_file(_filepath: &str)->Result<()>{
+    Ok(())
+}
+
+fn load_from_s3()->(){
```
jyn514, line 143: What is this for?

jyn514 (changes requested):

src/app.rs
nikhilmitrax: This is just to follow the convention defined by `bacon`, which is to check `src/` directory. The idea I had, I guess, of this MR was to _extend_ the usage to workspaces, but not necessarily change the `semantics` of what's being watched.

Open to suggestions, though, on what exactly to watch, which we could address in a separate MR.

src/app.rs
nikhilmitrax: The logic was quite flawed. For some reason, I was looking into the targets defined within the package, and trying to find the folder off of that.

Switched to directly using `parent()` of `pacakge.manifest_path`

nikhilmitrax:
@jyn514 Thank you very much for taking the time to review the code. 

I just pushed some commits to overhaul the workspace member finding method, as well as fix a bug where it was only watching the `src` directory of the workspace member, and not the manifest file.

Canop:
Sorry not to have reacted (or read) yet. I know this tackles a very important problem and needed feature but I'm very busy working on other parts and can't focus on this one right now.
Thank for your work.

Canop:
I'm incorporating without merge the interesting parts of this PR.

Discussion continues in issue #18

Issue #27, opened 2020-12-06
Support checking examples


Canop:
Right now bacon doesn't check examples.

A solution is to add a job for each example but it's not very convenient as it means you bot have to add the entry in `bacon.toml` and to choose it when launching bacon, and to select the example directory to have it watched.

So I'd like something better. And before implementing anything here, I'd like to decide on a design.

Ideas:

1) automatically add `--example [..]` when the path can be detected to be one of an example

2) have a special syntax for the command where `"<filename>"` would be expanded
```
[jobs.example]
command = ["cargo", "check", "--example", "<filename>", "--color", "always"]
```

PR #33, opened 2021-01-19
Change `check-all` to actually check all targets

jyn514:
Previously, it would only check tests and the main library. This includes binaries, examples, and benchmarks.

I think this counts as fixing https://github.com/Canop/bacon/issues/27, not sure if you had other plans for it.


Canop:
> not sure if you had other plans for it

I never found the `--all-targets` flag...

Issue #29, opened 2020-12-09
Automatically re-configure/re-run on updates to bacon.toml


cpick:
I've been testing various custom jobs by adding them to and editing "bacon.toml".

Currently my flow is:
1. Edit "bacon.toml".
2. Kill current `bacon`.
2. Re-run `bacon <custom_job>`.

I think it would be nice if `bacon` watched "bacon.toml" for updates and restarted itself to load the new configuration and try any changes.

If this sounds like a useful feature I wouldn't mind taking a try at implementing it.

Thanks for a great tool!

Canop:
I didn't think this would be needed but I have no objection.

If you don't find an obvious clean way to code it, I'll do it myself.

Canop:
Now, do we also want to track the preferences files ?

PR #35, opened 2021-01-19
Watch `bacon.toml` for changes

jyn514:
Closes https://github.com/Canop/bacon/issues/29.


Canop:
I may be missing something but I don't get the point here. Just getting a nofification the bacon.toml file changed won't reload it or takes the changes into account.

jyn514:
Oh oops, you're right. This is on the wrong track then.

Issue #34, opened 2021-01-19
`bacon --init` does nothing if bacon.toml already exists


jyn514:
I would expect it to give either an error or a warning. Instead it does nothing.

Steps to reproduce:
1. `touch bacon.toml`
2. `bacon --init`

PR #36, opened 2021-01-19
Print an error in --init if bacon.toml already exists

jyn514:
Example output:

```
Error: /home/joshua/src/rust/bacon/bacon.toml already exists, refusing to overwrite
```

Closes https://github.com/Canop/bacon/issues/34


Canop:
I'm changing it to keep allowing this construct:

    nvim $(bacon --init)

which opens the bacon.toml file, be it new or not

PR #37, opened 2021-02-03
Add README section about check-all option

PoignardAzur:
I wasn't aware of that option, and there was no way to discover it from the command line.

I only found it by looking through the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ or
 
     bacon ../broot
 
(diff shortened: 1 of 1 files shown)


Canop:
Good idea!

PoignardAzur:
I looked if there was a way to include it eg when the `--job` command is given without arguments, but `argh` doesn't look quite flexible enough. Oh well.

PR #38, opened 2021-02-08
Watch examples directory if it exists

SafariMonkey:
While the examples can be checked by `--check-all-targets` (#33) or `--examples` (which would be a nice addition to the default jobs), the examples directory is currently not watched by the watching mechanism. The change in this PR causes examples to be watched on my machine.

### DIFF
--- a/src/mission.rs
+++ b/src/mission.rs
@@ -112,6 +112,12 @@ impl Mission {
                     } else {
                         warn!("missing src dir: {:?}", src_dir);
                     }
+                    let examples_dir = item_path.join("examples");
+                    if examples_dir.exists() {
+                        directories_to_watch.push(examples_dir);
+                    } else {
+                        debug!("missing examples dir: {:?}", examples_dir);
+                    }
                 }
                 if item.manifest_path.exists() {
                     files_to_watch.push(item.manifest_path);


Canop:
examples and tests can be very big directories. I'm not sure that the approach of automatically following them is the right one.

There's a real need though. 

I was thinking about adding some paths to watch to the job description, something like
```
[jobs.test]
command = ["cargo", "test", "--color", "always"]
need_stdout = true
watch = ["tests"]
```
Your thoughts ?

SafariMonkey:
Thanks for responding so quickly!

That approach makes sense to me, and I wondered about the possibility before going for the smaller change for the PR. I was originally going to post an issue, but after seeing how small the minimal change would be, I thought I might as well - I figured you may well have a better suggestion. (For my future reference, would you generally prefer a possibly-wrong PR or an issue that simply lays out the problem in such a case? I honestly wasn't sure which was better, and would welcome feedback.)

If you're happy to add such a "watch directories" parameter, then that makes sense to me. I'm not sure if they would only be watched in the case of opening a package, but that's my thinking. It would be nice if this were preconfigured to reasonable values on the built-in jobs like `check-all` and may my suggested `examples`.

If you would like, I can have a go at implementing that. I don't mind either way, and you might find it faster to do it than go back and forth.

Canop:
There was no problem in your PR, especially as it wasn't big. But for future reference the best on my projects is to come to [Miaou](https://miaou.dystroy.org/3768) and have a chat.

If you really want to do a new PR and are ready for my unfair critics and pointless changes, go ahead. If you don't, I'll do it soon maybe this evening or the next one. In both cases an issue would be welcome.

PR #40, opened 2021-02-09
Allow specifying watch directories in bacon.toml

### LINKED ISSUE #39: Make watch directories configurable
Running `bacon -j check-all` reports errors in the examples, but if examples are changed, `bacon` does not rerun the check. I attempted to solve this in #38, but @Canop suggested that the following would be preferable:

```toml
[jobs.test]
command = ["cargo", "test", "--color", "always"]
need_stdout = true
watch = ["tests"]
```

Therefore, only the directories that the job depends on are watched.

One question I have: should `src` be included implicitly? Or should the user have to specify `"src", "tests"`?

SafariMonkey:
Allows specifying additional watch directories in a job. Closes #39.

I wasn't sure whether to combine the `src` with the iterator like this or just repeat the code block. Happy to change it.

Just to note, this _will_ `warn!` about missing benches, tests, and examples when you run `check-all`. If that should be `info!` or `debug!`, splitting the handling of `src` and other definitely makes sense, IMO.

Also, if I missed any doc updates that are needed, let me know.

### DIFF
--- a/src/mission.rs
+++ b/src/mission.rs
@@ -7,6 +7,7 @@ use {
         collections::HashSet,
         env,
         fs,
+        iter,
         path::PathBuf,
         process::Command,
     },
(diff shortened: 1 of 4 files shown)


Canop:
Good. I'll check this in a few days (sorry, big backlog)

src/mission.rs
```diff
@@ -106,11 +107,15 @@ impl Mission {
                     .parent()
                     .expect("parent of a target folder is a root folder");
                 if add_all_src {
-                    let src_dir = item_path.join("src");
-                    if src_dir.exists() {
-                        directories_to_watch.push(src_dir);
-                    } else {
-                        warn!("missing src dir: {:?}", src_dir);
+                    let src_watch_iter = iter::once("str");
```
Canop, line 110: "str" ?

(I'm fixing it, don't make a new commit)

SafariMonkey: Oh wow, sorry about that! It seems str was on my mind. Thanks for fixing it!

PR #45, opened 2021-04-30
add global config functionality

luciusmagn:


Canop:
Why do you add the dirs crate ? There's already another crate for that, used for example for the global settings.

luciusmagn:
Aaah, I didn't notice it, my bad, sorry

PR #46, opened 2021-06-02
Add a `bacon doc` task

cormacrelf:
I have been using this a bunch, it works really well. It even shows the unresolved `[doc_links::to_stuff]` warnings. `cargo doc` is blazing fast, so you can tweak docs and refresh the browser in under a second. I would highly recommend trying this out and including it by default.

Also something i've been using, which is awesome but probably not a default:

```toml
[jobs.docs-rs]
command = [
  "env",
  "RUSTDOCFLAGS=--cfg docsrs",
  "cargo",
  "+nightly",
  "doc",
  "--color", "always",
  "--no-deps",
  "--all-features",
]
```

Then you can preview with the same `--cfg docsrs` that you can configure via the Cargo.toml `[package.metadata.docs.rs]` options, and get the whole "this is only available with feature XXX" notices in your doc previews. (docs rs builds on nightly so the `doc_cfg` feature is available.) For anyone wanting to try, it's:

```toml
[package.metadata.docs.rs]
all-features = true # if you want
rustdoc-args = ["--cfg", "docsrs"]
```

And then

```rust
// crate root
#![cfg_attr(docsrs, feature(doc_cfg))]

// on items
#[cfg(feature = "feature-name")]
#[cfg_attr(docsrs, doc(cfg(feature = "feature-name")))]
struct S;
```

(I'm guessing this will be stabilised at some point, and then regular `bacon doc` would work if you remove the cfg_attr wrapping.)

### DIFF
--- a/defaults/default-bacon.toml
+++ b/defaults/default-bacon.toml
@@ -22,3 +22,7 @@ need_stdout = false
 command = ["cargo", "test", "--color", "always"]
 need_stdout = true
 watch = ["tests"]
+
+[jobs.doc]
+command = ["cargo", "doc", "--color", "always", "--no-deps"]
+need_stdout = false


Canop:
This is cool.

But there's a downside currently: it doesn't (in my tests) directly open the browser.

I can do `cargo doc --open & bacon doc` but it's clearly less ergonomic.

Does your experience match mine ? Do you have a workaround or an idea to improve this ?

cormacrelf:
Well, it isn't any less ergonomic than using `cargo doc --open` and then subsequently `cargo doc`. I suppose you're comparing it to a tool with both watch and open features, but using watch and open together is just as annoying in other ways and I try to avoid it if possible.

As for how I've been using `bacon doc`, I just `cargo doc --no-deps --open` once and (crucially) pin the tab, then finally run `bacon doc` and let it sit there for literally days at a time.

There is an ergonomic problem with `--open` and all equivalents across all the tools I've used, which is that:

- without a watch feature, you get a whole bunch of redundant tabs from hitting up+enter to recompute and catching a history item with --open in it. This is mostly irrelevant because bacon does watch files. And also because `bacon doc` is quick enough to type from scratch.
- if it can open and watch, you *still* get excess tabs because sometimes you have to restart. This is almost invariably often enough for it to be annoying. For `bacon doc` you might want to enable and disable features or whatever. It happens and it's annoying. **in my experience it is best if your last shell history item (= the watch command) does not have `--open` in it at all.**

So I find the only good way is as I described above. Pinning tabs really does help. 

Obviously there are some more options:

- add an --open flag to bacon proper, and then use it to add --open to any bacon command until the first successful execution only. (See below about failed doc builds)

- add a config file key of some kind, maybe a generic "run this first" command. I don't see any other obvious uses though.

- Add a full-on LiveReload-style bolt on. Too much work by far. Plus Rustdoc is apparently considering adding its own server for serving an archived output of the files in it without having to write so many files (esp slow on Windows). See [this blog post under "experimental archive format..."](https://quietmisdreavus.net/code/2019/02/19/rustdoc-in-2019/).

My best idea is saved for last:

- **Support `-- --passthru --flags` syntax.** That would be useful all over the place in bacon. And I find myself frequently trying it expecting it to work. It would let you tack on `-- --open` to bacon, let it go once, exit and delete the extra flag, run again. This would also be better because my PR has `--no-deps` basically because that should be the default for rustdoc anyway as deps take too long and `--no-deps` just links external references to docs.rs anyway. But `cargo doc --open` builds all dependencies and pollutes your documentation sidebar with heaps of irrelevant crates that you need to `cargo clean --doc` (?) or `rm target/doc` to get rid of. It's a lot of typing to avoid this with `cargo doc --no-deps --open`.

- Another generic improvement would be a flag or a config option to run the underlying command until it succeeds once and exit. This is useful because IIRC cargo doc won't open your browser if the doc build fails, so it's usually a run over and over again until it works deal. Maybe a flag would be good for more workflows? (Ooh, yes -- running bacon in a shell while loop, where you make sure it passes some tests and then run the next shell command in the loop. Combined with a `--wait` "don't run at all until I change a file" flag. `while true; do bacon check -u; cargo build --release; hyperfine --warmup 2 target/release/binary; done`. Noice.)

Back to the issue at hand, that looks like, with a flag:

```sh
# build docs until it opens
bacon doc -u -- --open
# watch from there
bacon doc
```

Or with a new config key:

```toml
[jobs.doc-open]
command = ["cargo", "doc", "--color", "always", "--no-deps", "--open"]
until-success = true
``` 

```sh
bacon doc-open
bacon doc
```

Canop:
This (welcome) discussion goes beyond the scope of a small PR/

What I propose:

- I merge your PR which already brings a useful common bacon mission
- you create an issue for passing args behind `--` (and from there I'll think about it)
- I create an issue to track the problem of the `--open` and we'll explore relevant strategies for that there

Do I miss something ?

cormacrelf:
Sounds good to me. Thanks for a great tool by the way. It's improved my life heaps.

Issue #10, opened 2020-11-04
bacon crashes on rustc


jyn514:
It runs `cargo check`, finishes seeing all the 'duplicate lang error's, then immediately exits with

```
Error: No space left on device (os error 28)
```

`df -h` reports 300 GB free, so I don't expect that to be the issue.

jyn514:
Here's an strace in case it helps: [bacon.log](https://github.com/Canop/bacon/files/5486659/bacon.log)

jyn514:
Here's `BACON_LOG=trace`:

```
09:05:10 [ INFO] Starting Bacon v0.2.3 with log level TRACE
09:05:10 [DEBUG] (1) bacon::cli: args: Args { version: false, summary: false, clippy: false, root: None }
09:05:10 [DEBUG] (1) bacon::app: root_dir: "/home/joshua/rustc"
09:05:10 [TRACE] (2) mio::poll: [/home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.7.5/src/poll.rs:474] registering event source with poller: token=Token(0), interests=READABLE
09:05:10 [TRACE] (2) mio::poll: [/home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.7.5/src/poll.rs:474] registering event source with poller: token=Token(1), interests=READABLE
09:05:10 [DEBUG] (1) bacon::report: starting cargo check
09:05:10 [DEBUG] (1) bacon::report: cargo check finished
09:05:10 [DEBUG] (1) bacon::report: status: ExitStatus(ExitStatus(25856))
09:05:10 [DEBUG] (1) bacon::report: report: 0 warnings and 1331 errors
09:05:10 [TRACE] (1) mio::poll: [/home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.22/src/poll.rs:787] registering with poller
09:05:10 [TRACE] (1) mio::poll: [/home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.22/src/poll.rs:787] registering with poller
09:05:10 [TRACE] (1) mio::poll: [/home/joshua/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.22/src/poll.rs:787] registering with poller
```

jyn514:
Maybe it's trying to write too much to the buffer?

Canop:
[...]

PR #53, opened 2021-09-11
Add a --workspace flag, to check entire workspace

Techcable:
This simply forwards to the --workspace flag
of cargo check and clippy.

I find this very useful when using `cargo check`
directly, so I figured it'd be nice to have here :)

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -57,7 +57,7 @@ pub struct Args {
     /// check all members in the workspace.
     ///
     /// Equivalent to `cargo check --workspace`
-    #[argh(switch)]
+    #[clap(long = "workspace")]
     pub workspace: bool,
 
     /// activate all available features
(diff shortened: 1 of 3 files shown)


Canop:
Hum... I just notice this PR I had forgotten...

Can I have a third-party review ?

Techcable:
I have rebased onto latest `main` commit, resolving the outstanding conflicts.

Canop:
I think this PR has been made obsolete by #10 : it's now possible to pass additional arguments after `--` which allows

`bacon -- --workspace`

Techcable:
> Think this PR has been made obsolete by https://github.com/Canop/bacon/issues/10 : it's now possible to pass additional arguments after -- which allows
>
> bacon -- --workspace

It appears you are correct. Since this only saves two characters (and two spaces), I'm not sure this is worth it anymore.

Thank you :)

PR #57, opened 2021-11-15
Add clippy-all to default-bacon.toml

rukai:
This is a real fairly subjective change so i'm happy to change things.

But I do feel strongly that the most aggressive config (running clippy on everything) is a highly desirable default to have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ or
 
     bacon clippy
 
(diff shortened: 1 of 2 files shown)


Canop:
I don't get your change in the README. Did you forget something after the copy-paste ?

rukai:
yes, sorry about that

PR #58, opened 2021-11-15
Sort --list-jobs output

rukai:
Its quite surprising when the order of the list changes each time its run.

### DIFF
--- a/src/list_jobs.rs
+++ b/src/list_jobs.rs
@@ -16,7 +16,9 @@ pub fn print_jobs(package_config: &PackageConfig) {
     default job: ${default_job}
     "#;
     let mut expander = OwningTemplateExpander::new();
-    for (name, job) in &package_config.jobs {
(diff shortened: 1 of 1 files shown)


Canop:
@rukai Do you think you'll do other PR soon ? Because I don't like to have the last release not in sync with the readme so I'd like to do a new version but I can wait if you think you'll be fixing other things.
You'd be welcome on the chat : https://miaou.dystroy.org/3768

rukai:
Nope that's it for now. I just happened to notice those 3 things in quick succession.

Stargateur:
hope that this isn't critical cause allocate a temporary string when you can do `.sort_by(|(a, _), (b, _)| a.cmp(b))`.

With itertools a one liner `package_config.jobs.iter().sorted_by(|(a, _), (b, _)| a.cmp(b))` allocate the vector internally.

Canop:
It's *really* non critical.

PR #60, opened 2021-11-16
Use cargo metadata to find cargo manifest

### LINKED ISSUE #59: Bacon fails on a workspace with inter-member dependencies on Windows
Reproduction steps:

1. Create a workspace manifest `./Cargo.toml`:

    ```toml
    [workspace]
    members = [
        "crate-a",
        "crate-b",
    ]
    ```

2. Create two member crates:

    ```
    cargo new --lib crate-a
    cargo new --lib crate-b
    ```

3. Make `crate-b` depend on `crate-a` (in `./crate-b/Cargo.toml`):

    ```toml
    [package]
    name = "crate-b"
    version = "0.1.0"
    edition = "2018"

    [dependencies]
    crate-a = { path = "../crate-a" }
    ```

4. Run `bacon` in workspace root, or in either of the crate directories. It fails with message:

    > Error: \`cargo metadata\` exited with an error: error: package collision in the lockfile: packages crate-a v0.1.0 (... dir1 ...) and crate-a v0.1.0 (... dir2 ...) are different, but only one can be written to lockfile unambiguously

Meanwhile `cargo metadata` and `cargo check` work correctly when ran in any of those 3 directories.

OS: Windows 10
`cargo --version`: `cargo 1.54.0 (5ae8d74b3 2021-06-22)`
`bacon --version`: `bacon 1.2.3`

jDomantas:
Use `cargo metadata` itself to find Cargo.toml instead of manually walking up the directory tree.

There is a small difference in behavior: now you can't run `bacon path/to/Cargo.toml` (and instead you have to do `bacon path/to`). It seems that previous behavior was accidentally allowed as it would treat the path like a directory and first check `path/to/Cargo.toml/Cargo.toml` for existence, followed by `path/to/Cargo.toml`.

The couple of `expect()`s should not fail unless I misunderstood cargo metadata docs.

Fixes #59.

### DIFF
--- a/src/mission.rs
+++ b/src/mission.rs
@@ -27,9 +27,20 @@ impl MissionLocation {
             .path
             .as_ref()
             .map_or_else(|| env::current_dir().unwrap(), PathBuf::from);
-        let metadata = MetadataCommand::new()
-            .current_dir(&intended_dir)
-            .exec()?;
+        let metadata = match MetadataCommand::new().current_dir(&intended_dir).exec() {
+            Ok(m) => m,
+            Err(cargo_metadata::Error::CargoMetadata { stderr }) if cargo_manifest_not_found(&stderr) => {
+                bail!(
+                    "Cargo.toml file not found.\n\
+                    bacon must be launched \n\
+                    * in a rust project directory\n\
+                    * or with a rust project directory given in argument\n\
+                    (a rust project directory contains a Cargo.toml file or has such parent)\n\
+                    "
+                );
+            }
+            Err(other) => bail!(other),
+        };
         let resolve = metadata
             .resolve
             .expect("cargo metadata should resolve workspace without --no-deps");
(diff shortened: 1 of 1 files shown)


Canop:
I'm OK for not allowing a path to the Cargo.toml file but there's a regression IMO.

Old and new behaviors when launched without path as argument and not in a Rust project:

image

Can you please restore the old error message ?

DomantasJ:
Restored the old message. Sadly `cargo metadata` does not emit structured error data, so I hope that string matching on stderr is good enough.

jDomantas:
Restored the old message. Sadly `cargo metadata` does not emit structured error data, so I hope that string matching on stderr is good enough.

Canop:
This is probably the right approach.
I'll have a lot of checks and tests to do before I can merge that, though. People with strangely structured workspaces are encouraged to test and tell me how it is.

Regarding the errors in `metadata`, a PR there would probably be a good idea too (it's still 0.x so this kind of breaking change should be OK).

Stargateur:
bacon is 1.x.y but I recommend to not follow semver on application, this is almost impossible cause the smallest change is a breaking change, thus major change should update the major, this should be considered as minor update. You will end with firefox and his 94 versions. That said it's your call.

Issue #78, opened 2022-04-06
Add option to forcefully cancel job on changes


nolanderc:
Currently, if the running job gets stuck in an infinite loop bacon  becomes unresponsive. This is easily reproduced by running `bacon test` in a project containing the following:

```rust
#[test]
fn infinite() {
    loop {}
}
```

Pressing Ctrl+C does not work, and neither does making changes to any source files (tested on macOS 12.0.1).

This is also relevant when running a server in a compile-edit-run cycle using `bacon run`, as these programs aren't expected to automatically terminate, but frequent rebuilds are useful.

This problem is solved in [`watchexec`](https://github.com/watchexec/watchexec) by using a flag (`--restart`) which forcefully terminates the running command if there are any changes to watched files.

Something similar could be added to bacon, either as an option in the configuration file, or as a command line flag.

nolanderc:
After some digging I have traced the problem to this line in the executor: https://github.com/Canop/bacon/blob/d256ebcb7e2244bcad335abccafb8ab84fc8aea2/src/executor.rs#L135

Basically, the executor is blocked on reading the next line from `cargo test`, but since the test is stuck, there is no output to be read and the executor is stuck. This leaves it unable to receive events from other threads. Working on a fix...

Canop:
I don't know why this issue wasn't closed...

PR #79, opened 2022-04-06
use asynchronous IO in the executor

nolanderc:
This is my second attempt at fixing #78. 

At first, I thought it would be enough to just call `Child::kill` upon receiving the stop signal in the executor (https://github.com/nolanderc/bacon/commit/07ab3184e2f870d75d36b26380375c0510409539). However, this didn't work, as it didn't fully close the `stderr` and `stdout` streams and thus kept the threads open in the background, still deadlocked.

This version uses `async` in order to cleanly cancel the currently running task. This does mean an extra dependency on `tokio`, but I've tried to keep the number of features down as much as reasonably possible.

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -91,7 +91,8 @@ impl Executor {
                                 Ok(child) => child,
                             };
 
-                            current_task = Some(tokio::spawn(                                execute_task(child, with_stdout, line_sender.clone())
+                            current_task = Some(tokio::spawn(
+                                execute_task(child, with_stdout, line_sender.clone())
                             ));
                         }
 
@@ -186,7 +187,11 @@ async fn execute_task(
             .take()
             .ok_or_else(|| anyhow!("child missing stdout"))?;
         let stdout_sender = line_sender.clone();
-        Some(stream_consumer(stdout, CommandStream::StdOut, stdout_sender))
+        Some(stream_consumer(
+            stdout,
+            CommandStream::StdOut,
+            stdout_sender,
+        ))
     } else {
         None
     };
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -178,8 +178,7 @@ async fn execute_task(
         .ok_or_else(|| anyhow!("child missing stderr"))?;
 
     let stderr_sender = line_sender.clone();
-    let stderr =
-        tokio::spawn(async move { consume_stream(stderr, CommandStream::StdErr, stderr_sender) });
+    let stderr = stream_consumer(stderr, CommandStream::StdErr, stderr_sender);
 
     let stdout = if with_stdout {
         let stdout = child
@@ -187,17 +186,16 @@ async fn execute_task(
             .take()
             .ok_or_else(|| anyhow!("child missing stdout"))?;
         let stdout_sender = line_sender.clone();
-        Some(tokio::spawn(async move {
-            consume_stream(stdout, CommandStream::StdOut, stdout_sender)
-        }))
+        Some(stream_consumer(stdout, CommandStream::StdOut, stdout_sender))
     } else {
         None
     };
 
-    // we use `unwrap` to bubble panics in the subtasks up the stack
-    stderr.await.unwrap().await?;
+    // either we wait on both stdout and stderr concurrently, or just stderr.
     if let Some(stdout) = stdout {
-        stdout.await.unwrap().await?;
+        tokio::try_join!(stdout, stderr)?;
+    } else {
+        stderr.await?;
     }
 
     let status = match child.wait().await {
@@ -209,7 +207,7 @@ async fn execute_task(
 }
 
 /// Send all lines in the given stream to the sender.
-async fn consume_stream(
+async fn stream_consumer(
     stream: impl AsyncRead + Unpin,
     origin: CommandStream,
     line_sender: LineSender,
@@ -218,9 +216,9 @@ async fn consume_stream(
 
     while let Some(line) = lines.next_line().await? {
         let response = CommandExecInfo::Line(CommandOutputLine {
-                content: TLine::from_tty(&line),
-                origin,
-            });
+            content: TLine::from_tty(&line),
+            origin,
+        });
         if let Err(_) = line_sender.send(response) {
             return Err(anyhow!("channel closed"));
         }
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -1,12 +1,17 @@
-use {
-    crate::*,
-    anyhow::Result,
-    crossbeam::channel::{bounded, select, unbounded, Receiver, Sender},
-    std::{
-        io::{BufRead, BufReader},
-        process::Stdio,
-        thread,
+use crate::*;
+use anyhow::{anyhow, Context, Result};
+use std::{
+    process::{ExitStatus, Stdio},
+    thread,
+};
+use tokio::{
+    io::{AsyncBufReadExt, AsyncRead, BufReader},
+    process::{Child, Command},
+    sync::{
+        mpsc::{channel, Sender},
+        oneshot,
     },
+    task::JoinHandle,
 };
 
 /// an executor calling a cargo (or similar) command in a separate
@@ -14,9 +19,9 @@ use {
 /// and finishing by None.
 /// Channel sizes are designed to avoid useless computations.
 pub struct Executor {
-    pub line_receiver: Receiver<CommandExecInfo>,
+    pub line_receiver: crossbeam::channel::Receiver<CommandExecInfo>,
     task_sender: Sender<Task>,
-    stop_sender: Sender<()>, // signal for stopping the thread
+    stop_sender: oneshot::Sender<()>, // signal for stopping the thread
     thread: thread::JoinHandle<()>,
 }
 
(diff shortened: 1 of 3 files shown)


Canop:
Thanks.
It will take some time to test that, and I'll also see whether there's a simple way to avoid the tokio dependency, so don't expect a merge before the week-end.

Canop:
Just tried with
```
fn main() {
    println!("starting...");
    loop {}
    println!("done");
}
```
The first `println!` isn't displayed

nolanderc:
Pushed a new fixed which I have confirmed fixed the issue: see the commit message for details (my understanding of `tokio::spawn` is obviously not as good as I thought it was).

nolanderc:
Something else I just noticed with the sample above: while it is stuck in the loop, and a change is made to the code the old output is not cleared. This is probably best handled in `app.rs`, but I'm not too familiar with the internals (maybe we are missing a call to `state.set_result(...)`?

nolanderc:
After further investigations: doing the same test (running the above snippet and re-writing the file) the line `starting...` is shown the first time bacon is run, but after the re-write the line is not shown. Using logging statements I have confirmed that the line is sent from the executor to the `App`, but the line is not shown. Must be some bug in `App`...

Canop:
I'll have a look ASAP

Canop:
`br ~` should be instant. You must have specific options.

I notice for example in you screenshot the filesystem bar at the top of the screen. Displaying it may be slow if you have among your mounts a slow one, or a disconnected one.

Try launch broot without options (just temporarily remove the conf.hjson file) to check the difference.

src/executor.rs
```diff
@@ -25,161 +30,90 @@ pub struct Task {
...
                                 }
+                                Ok(child) => child,
                             };
-                            if let Err(e) = line_sender.send(r) {
-                                debug!("error when sending stderr line: {}", e);
-                                break;
-                            }
-                            buf.clear();
+
+                            current_task = Some(tokio::spawn(                                execute_task(child, with_stdout, line_sender.clone())
```
Stargateur, line 94: SPAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACE ?!?

src/executor.rs
```diff
@@ -199,3 +133,96 @@ impl Executor {
...
+    task: &mut Option<JoinHandle<Result<Option<ExitStatus>>>>,
+) -> Result<Option<ExitStatus>> {
+    match task {
+        Some(handle) => handle.await.unwrap(),
+        None => match AlwaysPending.await {},
+    }
+}
+
+/// A future that will never resolve
+struct AlwaysPending;
```
Stargateur, line 147: https://doc.rust-lang.org/std/future/fn.pending.html already do this

src/executor.rs
```diff
@@ -199,3 +133,96 @@ impl Executor {
...
+    } else {
+        None
+    };
+
+    // either we wait on both stdout and stderr concurrently, or just stderr.
+    if let Some(stdout) = stdout {
+        tokio::try_join!(stdout, stderr)?;
+    } else {
+        stderr.await?;
+    }
```
Stargateur, line 199: this should be one if

src/executor.rs
```diff
@@ -25,161 +30,90 @@ pub struct Task {
...
-                                )) {
-                                    debug!("error when sending stderr error: {}", e);
-                                    break;
-                                }
-                                continue;
+            // start a runtime to manage the executor
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_io()
+                .build()
+                .unwrap();
```
Stargateur, line 59: I believe it would be better if we use tokio in bacon to init it in main.

src/executor.rs
```diff
@@ -199,3 +133,96 @@ impl Executor {
...
+    let mut lines = BufReader::new(stream).lines();
+
+    while let Some(line) = lines.next_line().await? {
+        let response = CommandExecInfo::Line(CommandOutputLine {
+            content: TLine::from_tty(&line),
+            origin,
+        });
+        if let Err(_) = line_sender.send(response) {
+            return Err(anyhow!("channel closed"));
+        }
```
Stargateur, line 224: this is block call in async function, this look bad.

src/executor.rs
```diff
@@ -25,161 +30,90 @@ pub struct Task {
...
     /// If `with_stdout` capture and send also its stdout.
     pub fn new(mission: &Mission) -> Result<Self> {
-        let mut command = mission.get_command();
+        let mut command = Command::from(mission.get_command());
         let with_stdout = mission.need_stdout();
-        let (task_sender, task_receiver) = bounded::<Task>(1);
-        let (stop_sender, stop_receiver) = bounded(0);
-        let (line_sender, line_receiver) = unbounded();
+        let (task_sender, mut task_receiver) = channel::<Task>(1);
+        let (stop_sender, mut stop_receiver) = oneshot::channel();
```
Stargateur, line 43: unclear where is handle the signal, and tokio have tool that already return a stream https://docs.rs/tokio/latest/tokio/signal/index.html

src/executor.rs
```diff
@@ -25,161 +30,90 @@ pub struct Task {
...
+                let mut current_task: Option<tokio::task::JoinHandle<_>> = None;
+
+                loop {
+                    tokio::select! {
+                        _ = &mut stop_receiver => break,
+
+                        // wait for the next task
+                        task = task_receiver.recv() => {
+                            if let Some(old) = current_task.take() {
+                                old.abort();
```
Stargateur, line 72: where is define this ? `Task` is define in this file but abort is nowhere to be found

src/executor.rs
```diff
@@ -25,161 +30,90 @@ pub struct Task {
...
+
+                            let child = match start_task(task, &mut command) {
+                                Err(e) => {
+                                    let response = CommandExecInfo::Error(
+                                        format!("failed to start task: {}", e)
+                                    );
+                                    match line_sender.send(response) {
+                                        Err(_) => break,
+                                        _ => continue,
                                     }
```
Stargateur, line 89: block call in async

Stargateur (commented):
overall, I found the code is unclear (the old code too), this kind of code is critical and error prone I think this require refactoring, more comment and overall improve clarity.

src/executor.rs
nolanderc: This is a tokio task (`abort` is in `tokio::task::JoinHandle`)

src/executor.rs
nolanderc: Not sure what you are reffering to, could you be more specific?

Canop: There's a formatting problem. But it doesn't matter much, I'll reformat if there's nothing more important.

nolanderc: Never mind, I see it now... had to scroll more to the right. Strange that `rustfmt` left it that way...

Canop:
note: I'm not letting this die, I'm just cautious before adding such a dependency. I'm testing this branch at work since a few days.

src/executor.rs
nolanderc: Nice, didn't know that!

src/executor.rs
nolanderc: Probably, but I didn't want to expand the scope of this PR more than necessary, so I kept it confined to this single file.

src/executor.rs
nolanderc: ctrl+c is handled elsewhere `src/keybindings.rs` seems to define a handler for it (later used in `app.rs`). This "signal" is then used by the app in the `Executor::die` function to signal that the child process should be terminated.

src/executor.rs
nolanderc: Agreed, I wanted to keep using the `std` channels so that other code wasn't affected. But we could probably swap to [`tokio::sync::mpsc`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.blocking_recv) where relevant instead.

src/executor.rs
nolanderc: [See above](https://github.com/Canop/bacon/pull/79#discussion_r851250263)

src/executor.rs
nolanderc: It could be, but I wanted to keep stream "creation" and awaiting separate for clarity.

Stargateur: this doesn't work on that case you are adding complexity, less branch the better.

src/executor.rs
Stargateur: MMM it seem send is non blocking so all good https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#method.send

src/executor.rs
Stargateur: if possible I would use tokio to handle signal too avoid this extra channel would be nice

src/executor.rs
Stargateur: It's unclear why you use a task and not stock directly the Command and use `kill()`. I unsure this correctly clean the child process.

nolanderc: It will, due to the call to `kill_on_drop` on the command builder (https://github.com/Canop/bacon/blob/fd714b62e4760abac8947c7a6fc5886fc5cd0f72/src/executor.rs#L164).

src/executor.rs
nolanderc: This would require restructuring of the rest of bacon, which feels outside the scope of this PR.

Stargateur: Unfortunately, your PR is a big change to bacon, introduce tokio/async is not a small thing it's a major change for bacon.

PR #80, opened 2022-04-25
feat: Use cargo aliases through `--alias <alias>`

### LINKED ISSUE #77: Detect and use cargo aliases as jobs
I have an alias in `.cargo/config` that is `clw = "clippy --target x86_64-pc-windows-msvc --all-targets"`.

`bacon -j clw` finds nothing and I don't want to duplicate all my aliases into a bacon config file each time I add aliases for a project

Maybe with a new flag `-a / --alias` to make it clear it's not using a normal bacon job ?

poliorcetics:
- Adds an `--alias <alias>` flag to run `cargo <alias>` through bacon
  - `--color always` is added to the command
  - Watched directories are not supported
  - "Needs stdout" is not supported
  - Running a command on success is not supported
- Adds `-no-default-alias-args` to remove the default behaviour of adding `--color always`

This is an MVP, their could be more work done to support the missing features, for example
by extending to `bacon.toml` to support an `is_alias` flag on jobs or maybe an `[alias]` section.

Fixes #77

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -9,7 +9,6 @@ use {
 /// Documentation at <https://dystroy.org/bacon>
 #[clap(version, about)]
 pub struct Args {
-
     /// print the path to the prefs file, create it if it doesn't exist
     #[clap(long = "prefs")]
     pub prefs: bool,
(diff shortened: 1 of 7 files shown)


Canop:
While your PR was helpful, I finally decided against using a dedicated launch argument.
Instead it's now possible to refer to a job like `alias:my_alias`. An upside is that this syntax can also be used in default or in key bindings.

Issue #87, opened 2022-09-17
v1.2.5 and up fails to install on windows


Fluhzar:
All errors come from attempting to install `bacon` through `cmd` with the toolchain `stable-x86_64-pc-windows-msvc` via `cargo install bacon@{version}` where `{version}` is any of the version from `1.2.5` through `2.2.2`.

For the most recent versions, v2.2.1 and v2.2.2, the only error is this:

```
error[E0599]: no method named `enable_io` found for struct `tokio::runtime::Builder` in the current scope
  --> C:\Users\------\.cargo\registry\src\github.com-1ecc6299db9ec823\bacon-2.2.2\src\executor.rs:57:18
   |
57 |                 .enable_io()
   |                  ^^^^^^^^^
   |                  |
   |                  private field, not a method
   |                  help: there is an associated function with a similar name: `enable_all`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `bacon` due to previous error
```

Since the changes for v2.2.1 resolved all but the one above error I don't think it's particularly useful to share these errors, but if desired the errors I encounter when attempting to install any version of `bacon` from v1.2.5 through v2.2.0 are along the lines of:

<details>
<summary>v2.2.0 errors</summary>

<pre>
error[E0308]: mismatched types
  --&gt; C:\Users\------\.cargo\registry\src\github.com-1ecc6299db9ec823\bacon-2.2.0\src\app.rs:47:21
   |
46 |                 match user_event?.event {
   |                       ----------------- this expression has type `crokey::crossterm::event::Event`
47 |                     Event::Resize(mut width, mut height) =&gt; {
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `crokey::crossterm::event::Event`, found enum `crossterm::event::Event`
   |
   = note: perhaps two different versions of crate `crossterm` are being used?

error[E0308]: mismatched types
  --&gt; C:\Users\------\.cargo\registry\src\github.com-1ecc6299db9ec823\bacon-2.2.0\src\app.rs:57:21
   |
46 |                 match user_event?.event {
   |                       ----------------- this expression has type `crokey::crossterm::event::Event`
...
57 |                     Event::Key(key_event) =&gt; {
   |                     ^^^^^^^^^^^^^^^^^^^^^ expected enum `crokey::crossterm::event::Event`, found enum `crossterm::event::Event`
   |
   = note: perhaps two different versions of crate `crossterm` are being used?

[...]

PR #88, opened 2022-09-17
Add net feature of tokio for enable_io

Stargateur:
Closes https://github.com/Canop/bacon/issues/87

Look bacon run ok on windows with this.


Canop:
@Stargateur Did you check it fixes 87 on Windows ?

Stargateur:
Yes, it's compile and look ok. run some command, only thing weird is I can't open help menu with '?' but 'h' work, I think it's may be azerty keyboard thing. But all job work, clippy test etc

Canop:
The '?' key problem must be unrelated

Canop:
Thanks a lot @Stargateur

Issue #86, opened 2022-09-15
Bacon eats the last line of output when using `bacon run`


BLucky-gh:
When running `cargo run` I get this (cropped to the end)
```rust
HOST: localhost
DATABASE_URL: postgres://dev:devpwd@localhost/booru_dev
``` 
(the prompt is at the very next line with no extra newline except the one added by the prompt)
![image](https://user-images.githubusercontent.com/41377306/190406236-591fea70-873d-4184-aa6e-2c8eb0ecbc67.png)

but when running `bacon run` i get this(I have scrolled to the very bottom and scrolling more doesn't do anything, even though the scrollbar isn't at the very bottom):
```
HOST: localhost
Hit q to quit, w to wrap lines, h for help
```
![image](https://user-images.githubusercontent.com/41377306/190405707-9329c749-4a58-4d66-a212-4373b3d32937.png)

if it makes any difference: I'm running on MacOS and the terminal emulator is `kitty`. `bacon --version` returns `bacon 2.2.2`

jamesrampton:
Same here. Tested on MacOS in kitty and iTerm2. `bacon --version` > `bacon 2.2.5`

The behaviour can be reproduced if the content fills the window, but it's easier to see in reverse mode:

Normal:
<img width="1440" alt="Screenshot 2022-10-12 at 08 57 41" src="https://user-images.githubusercontent.com/1833713/195284493-da24b9e0-8d7c-4014-bac2-55323d98bb41.png">

Reverse:
<img width="1440" alt="Screenshot 2022-10-12 at 08 56 44" src="https://user-images.githubusercontent.com/1833713/195284555-4e5411df-12d4-4273-8b50-25768f41fa09.png">

sdrap:
I also meet this behavior when the output needs to scroll (last line is ignored).

System: ubuntu 22.04
term: default term, kitty, alacrity

with or without tmux

when the output fits without scrolling then no pb.

Canop:
Just a note that I don't reproduce the bug, even in reverse.

jamesrampton:
I can reproduce in Kitty, iTerm2 and Terminal.app with and without tmux. Admittedly only on MacOS Ventura (13.1).

I'd be interested to know what your environment is.

jamesrampton:
I've narrowed this down to the extra "computing" line that's added for long-running processes like web servers.

When this line is present, the last line of output is eaten.

[...]

PR #91, opened 2022-10-07
fix some cases of bad scrollbar

Canop:
They were related to wrapping not taken into account when computing content height


UncleScientist:
will this fix #86 as well?

Canop:
@UncleScientist Probably but I don't reproduce #86 so I can't be sure

PR #97, opened 2022-12-17
Fix typo in export locations doc

Jasha10:


Jasha10 (commented):
This PR fixes a typo in the docs: it is bacon that needs to be launched with the `-e` argument, not neovim.

Canop:
Oh thanks. A confusing error indeed.

PR #98, opened 2022-12-29
Check if file is ignored according to git

### LINKED ISSUE #32: Ignore some files
Some editors create temporary or lock files in the same directory as the file being edited. For example vi/vim will create a file named `.FILENAME.swp` when the file is first opened for editing, and emacs will create a link named `.#FILENAME` when a file open for editing has unsaved changes. When these files are created, they trigger an unnecessary extra run of the current job.

See https://github.com/watchexec/watchexec/blob/35e39afb95d5c66b41f843ce8850ee8da37a5516/src/cli.rs#L265 for a good starting point on which file patterns can be safely ignored. Obviously the `*.py?` patterns don't apply here.

Canop:
Doesn't launch the job if the file is ignored

There's a configuration parameter at the job level to disable this filtering.
Fix #32

### DIFF
--- a/src/ignorer.rs
+++ b/src/ignorer.rs
@@ -16,6 +16,10 @@ pub struct Ignorer {
 
 impl Ignorer {
 
+    /// Create an Ignorer from any directory path: the closest
+    /// surrounding git repository will be found (if there's one)
+    /// and its gitignore rules used.
+    ///
     /// root_path is assumed to exist and be a directory
     pub(crate) fn new(root_path: &Path) -> Result<Self> {
         let repo = git::discover(root_path)?;
@@ -25,15 +29,10 @@ impl Ignorer {
     /// Tell whether the given path is excluded according to
     /// either the global gitignore rules or the ones of the repository
     pub fn excludes(&mut self, file_path: &Path) -> Result<bool> {
-        // TODO this code does work but is wasteful (creating a cache
-        // for each file, for example). I hope this should get better
-        // when gitoxide matures
-
         let worktree = self.repo.worktree().context("a worktree should exist")?;
         let index = worktree.index()?;
 
-        // there doesn't seem to be any public API for looking at "excludes" without caching
-        // so we create a cache
+        // the "Cache" is the structure allowing checking exclusion
         let mut cache = worktree.excludes(&index, None)?;
 
         // cache.at_path panics if not provided a path relative
@@ -46,7 +45,7 @@ impl Ignorer {
         };
 
         // cache.at_path panics if the relative path is empty, so
-        // we must check it
+        // we must check that
         if relative_path.as_os_str().is_empty() {
             return Ok(true);
         };
--- a/src/ignorer.rs
+++ b/src/ignorer.rs
@@ -1,0 +1,71 @@
+use {
+    anyhow::{Context, Result},
+    git_repository::{
+        self as git,
+        prelude::FindExt,
+        Repository,
+    },
+    std::path::{Path, PathBuf},
+};
+
+/// An object able to tell whether a file is excluded
+/// by gitignore rules
+pub struct Ignorer {
+    repo: Repository,
+}
+
+impl Ignorer {
+
+    /// root_path is assumed to exist and be a directory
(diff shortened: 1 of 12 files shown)


Canop:
To test this and have more insight, launch it with info level log:

    BACON_LOG=info bacon

src/ignorer.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn excludes(&mut self, file_path: &Path) -> Result<bool> {
+        // TODO this code does work but is wasteful (creating a cache
+        // for each file, for example). I hope this should get better
+        // when gitoxide matures
+
+        let worktree = self.repo.worktree().context("a worktree should exist")?;
+        let index = worktree.index()?;
+
+        // there doesn't seem to be any public API for looking at "excludes" without caching
+        // so we create a cache
```
Byron, line 36: Cache is probably a misnomer here as it doesn't cache anything - it merely builds state to be able to perform these lookups and be fast with non-random input ordered according to the git index.
I think I ran out of words as it contains an ignore `stack` internally, and this is adding even more information to bring everything together.

src/ignorer.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// root_path is assumed to exist and be a directory
+    pub(crate) fn new(root_path: &Path) -> Result<Self> {
+        let repo = git::discover(root_path)?;
+        Ok(Self { repo })
+    }
+
+    /// Tell whether the given path is excluded according to
+    /// either the global gitignore rules or the ones of the repository
+    pub fn excludes(&mut self, file_path: &Path) -> Result<bool> {
```
Byron, line 27: I think as it stands, each path will trigger reading all gitignore files. They are indeed held in the `excludes()` data structure and ideally it is kept. I see how this isn't possible right now, and believe that the current reference is likely a premature optimization rather than a necessity. This will change for sure - and it's done in the latest main, which should greatly improve performance as the cache can actually reuse state if it's kept around.

Canop: > the current reference is likely a premature optimization rather than a necessity

I feel you. I fell in the same trap in my first libs too, and it's a pain to fix.

TBH checking whether a file is excluded takes less than 1 ms right now and that's fine for bacon.

Byron: I will think about this - the number one usage of lifetimes in structs is _platforms_, they add some data around an operation and perform it, keeping a reference to their originating `Repository`. As long as these are basically free, I think they can be created on demand with the `Repository` being cloned to where it is needed - that's the intent.

If they aren't free though, like the `Cache` here, I think it's good advice to rather clone the `Repository` into it to make it standalone, or do whatever else it takes. I think some useful rule emerges from this experience and I will put it into words in `DEVELOPMENT.md` to make it official.

Canop: BTW @Byron I don't know if you may find this interesting, but I've also implemented a gitignoring stack (stacking the parsed gitignore files as I imagine you do): https://github.com/Canop/broot/blob/main/src/git/ignore.rs#L155

This was done for broot with very specific performance concerns (breathfirst tree diving).
The reasons I didn't take that for bacon were
1. I wanted to try gitoxide for other programs of mine
2. I didn't want to implement myself looking for the global gitignore rules (I use git2 in broot for that)

Byron: Thanks for sharing! I love the perceived simplicity of the git-ignore implementation, it fits in 240 lines after all :)!

With the `Cache` type unchained from the lifetime in `main` you would now be able to reuse it for each lookup and that should yield much better performance.

From a correctness point of view, it's probably (hopefully)  a good idea to use `gitoxide` even if the performance is _just_ similar, as I tried my best to validate the implementation against git with many many test cases. Of course I hope you won't take my word for it and validate it yourself, `gitoxide` strives to yield the same results as `git`.

Thus, I hope you will end up using `gitoxide` in more of your projects and if there is anything preventing that, I'd love to know to get a chance to fix it.

Canop: gitignoring is only part of using git, and I definitely don't want to expand into building a general git crate while there's already an ambitious project, so it's my clear intention to try and use gitoxide ;)

PR #106, opened 2023-01-18
Add a "rerun" internal to force a new execution

### LINKED ISSUE #105: [Feature request] key shortcut to force a re-run
Hey, big user of your excellent pragmatic tool here 😄 

One thing that sometimes happens, is that `bacon` `watch` heuristics may not be 100% accurate with the real file-dependency involved in a complex project (_e.g._, imagine a `build.rs` depending on some weirdly-located file, or perhaps more realistically, a project depending on another one located at some other folder).

In that case, when I (hopefully rarely) update a file unbeknownst to `bacon`, **I have to explicitly press `q` followed by `↑` and `↩`** to force a `bacon` rerun.

It would be awesome if, akin to `q`, **we had some shortcut to trigger a rerun of the registered command**. Thanks! 🙏

Canop:
This can be used when a file change wasn't detected, when a "cargo run" execution might be interesting again, or for whatever reason user find.
It's not bound by default but can be bound by adding this line to a `[keybindings]` of a config file:

```TOML
f5 = "rerun"
```

Fix #105


danielhenrymantilla (approved):
Just tested it and it works exactly as I would have imagined, thanks for the super-fast implementation!

Issue #108, opened 2023-02-04
Warnings prevent test results from appearing


zolrath:
Hello!
Really enjoying using bacon, the usability of it has been great aside from one issue that keeps popping up for me;

Various warnings will prevent test results from appearing in `bacon`, even though the tests are being run and do appear if `cargo test` is run.

## With no warnings:
![image](https://user-images.githubusercontent.com/454563/216731589-589cde51-7850-4047-9bb0-61489e4b5811.png)

## With warning for unused function:
![image](https://user-images.githubusercontent.com/454563/216731622-81e3dc51-e2c6-471e-ad63-306d3e9c6f53.png)

## Cargo test
![image](https://user-images.githubusercontent.com/454563/216731656-2d18a120-5886-442e-ba32-f74da502d2e6.png)

# Expectation
I'd expect to still be able to see my test results.

Thank you very much!

Canop:
The tests are ran, and you would see the failures before the warnings:

![image](https://user-images.githubusercontent.com/617006/216752896-6b835fa4-3b5a-4b09-b5bc-d9cd06394f46.png)

I think it's what most people want: all failures and warnings, but less cluttering.

If you also want to the "running x tests" message with the list of tests below, then you may add `allow_warnings=true` in the job definition in either the `bacon.toml` or `bacon/prefs.toml` file:

```TOML
[jobs.test]
command = ["cargo", "test", "--color", "always"]
need_stdout = true
allow_warnings = true
watch = ["tests"]
```

Then you get this:

![image](https://user-images.githubusercontent.com/617006/216753048-5fba2b9d-74b5-42da-b498-cf631fee7ae7.png)

zolrath:
I suppose keeping the header saying `[test] [pass!] [1 warning]` instead of `[pass]` being entirely swallowed up in the header section would help it feel like the tests are still being run and make the header consistent across all combinations of statuses.

## Currently
#### Passing
`[test] [pass!]`

#### Passing + Warning
`[test] [1 warning]`

#### Failing
`[test] [1 fail]`

#### Failing + Warning
`[test] [1 fail] [1 warning]`

## Proposal
#### Passing + Warning
`[test] [pass!] [1 warning]`

zolrath:
Closed by #109

PR #109, opened 2023-02-12
Show pass with warnings/hide pass when 0 passing tests

zolrath:
Hello! This is in response to #108 with some code examples instead of just an ASCII representation of the badges.

The intent here is to make showing pass/fail more consistent, though while making these modifications I realized that `pass!` will currently show up in a project with 0 tests. 

To correct this, we only show the pass! badge when there is at least one passing test.

## Fail
fail

## Pass
pass

## No Tests
image

# Unnecessary Passes
Without this patch there are a number of places where `[pass!]` shows up erroneously.
image
image
This can be misleading as you may have failing tests, but when you are in clippy mode you are presented with a false `pass!` badge even if you have failing tests as clippy does not run them.

This PR removes pass from displaying in these instances, as it will only show up if successful tests have been performed.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -364,12 +364,11 @@ impl<'s> AppState<'s> {
             }
             if stats.test_fails > 0 {
                 t_line.add_badge(TString::num_badge(stats.test_fails, "fail", 235, 208));
+            } else if stats.passed_tests > 0 {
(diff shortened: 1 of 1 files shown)


Canop:
Sorry to be silent. I've seen this PR but didn't have time yet to seriously look at it

Canop:
It seems good.

PR #114, opened 2023-02-22
Update README.md

c-git:
Propose wording changing to make the text flow a bit better (imo).

- Because the action is not ongoing running is not the correct conjugation of the verb in that sentence.
- For "side to" it could also have been "on the side of" but I think alongside better captured what I think you were trying to say.
- Usually already means most of the time so the most seemed redundant but that one is a matter of taste.

Feel free to accept or reject any as you see fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 **bacon** is a background rust code checker.
 
-It's designed for minimal interaction so that you can just let it running, side to your editor, and be notified of warnings, errors, or test failures in your Rust code.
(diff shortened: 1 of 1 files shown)


Canop:
Seems better, thanks.

Canop:
Can you also change the website: https://dystroy.org/bacon/ ?

The file is at https://github.com/Canop/bacon/blob/main/website/docs/index.md

c-git:
Sorry got tied up with other stuff. Hope you got that sorted.

PR #120, opened 2023-03-13
Use different CSI codes depending on the platform

### LINKED ISSUE #70: Warnings & errors not recognized on some Windows terminals
I'm using Bacon 2.0.0 on Windows. If I run `bacon check` or `bacon clippy` and the program has errors, everything works as intended (even tho the shortcuts for wrapping lines and summary mode are not available, not sure if it is intended when there are errors), and the output looks something like this:
image

Instead, if only warnings are present, I can see bacon listing them for a fraction of a second, but then the screen turns to this:
image

https://user-images.githubusercontent.com/50493113/154508472-6a549e44-edeb-4fc2-a184-ffa332a42464.mp4

Basically, only the number of warnings is showed and not the actual warnings themselves, and I hope you can see from the video that for a moment a longer output is showed. I've tested this on a newly created project. 

Rust version is 1.58.1.

crillon:
Hello!

This fixes #70 

I do not know why, but on Windows, other CSI codes appears in the 'lines', so LineAnalysis returns garbage. 
This fix changes some CSI codes in tty, and also slightly changes the conditions by which the line type is determined in LineAnalysis on Windows

### DIFF
--- a/src/line_analysis.rs
+++ b/src/line_analysis.rs
@@ -3,6 +3,11 @@ use {
     lazy_regex::*,
 };
 
+#[cfg(not(windows))]
+const CSI_ERROR_BODY: &str = CSI_BOLD;
+#[cfg(windows)]
+const CSI_ERROR_BODY: &str = CSI_BOLD_WHITE;
+
 /// result of the "parsing" of the line
 #[derive(Debug, Clone)]
 pub struct LineAnalysis {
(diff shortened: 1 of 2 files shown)


Canop:
Interesting. Unfortunately I have no way to check it's right. Can another Windows user test that ?

Canop:
Are we about sure the proper check is `#[cfg(windows)]` (i.e. rustc doesn't look for the kind of terminal or shell) ?

crillon:
I think yes. I have tested this on different terminals and shells. And it seems that the problem appears in Windows regardless of the terminal/shell.

Also I choose to use `#[cfg(windows)]` because it has already been used here:
https://github.com/Canop/bacon/blob/78f982261085f287e59e3f792bbdb7c956ba872c/src/app.rs#L69-L73

I'm not sure about `#[cfg(not(windows))]`, maybe `#[cfg(unix)]` is better?

Canop:
> I'm not sure about #[cfg(not(windows))], maybe #[cfg(unix)] is better?

`#[cfg(not(windows))]` looks fine unless we find a better signature of when the CSI are different.

Canop:
I'll try to release a new version tomorrow

Issue #153, opened 2023-07-27
Can't scroll using mouse in a terminal ?


Adham-A:
Hello,
I can't seem to scroll using my mouse in a terminal, I tried outside and inside vscode but without success. Is it even possible or is it an issue on my computer ?

Canop:
vscode ? Is that issue related to bacon ?

If it is, then yes, you should be able to use the mouse wheel to scroll the report in bacon.

Adham-A:
### On Linux  
- Scrolling fires the "Up" & "Down" events, same as using those buttons on the keyboard.

### On Windows 
- No events are fired, but there is supposed to be.

Investigations :
I tried getting Mouse Events instead of key events inside the project with no luck, I then tried in an empty project with only `termimad` and I didn't succeed either. I then directly tried with crossterm and it worked, firing `ScrollDown` & `ScrollUp`

After some investigation you have to :

```rust
   // Have to do this part because of crossterm
    let mut stdout = io::stdout();
    execute!(
        stdout,
        EnableMouseCapture, // termimad::crossterm::event::EnableMouseCapture
    )?;
    // Have to do this part because of crossterm`
```

Canop:
Thanks for the investigation. I should have rememembered this myself as I had the problem before in broot.

Regarding your code, I'll comment on the PR

Canop:
Now fixed in 2.13.0

willhansen:
(On linux)

Works in normal terminals, but not in tmux.  I think tmux detects if the program captures mouse events, and forwards the mouse events to the program only if so.  

Probably related to the mentioned `EnableMouseCapture` thing.

PR #156, opened 2023-08-21
fix: :bug: Fix scroll in terminals on Windows

Adham-A:
#153

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -9,6 +9,13 @@ use {
     termimad::crossterm::event::Event,
     termimad::EventSource,
 };
+#[cfg(windows)]
+use termimad::crossterm::event::{
+    MouseEventKind,
+    KeyCode,
+    KeyEvent,
+    KeyModifiers,
+};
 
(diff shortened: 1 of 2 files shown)


Canop:
When you enable mouse capture in a program, you must disable it on end, or the terminal ends in a bad state.

Similar modifications are done then removed in `cli.rs` (look at `w.queue(cursor::Hide)?;` and `w.queue(cursor::Show)?;`). That's where you must move mouse enabling and disabling (and use `queue`, not `execute`, as for others).

Adham-A:
I now enable and disable in `cli.rs`,
I added `#[cfg(windows)]` in enabling and disabling, as well as in imports, since it showed unused warnings on Linux,

Something else to change/refactor ?

Canop:
> Something else to change/refactor ?

Nothing that I see right now. If you tested it seriously on Windows I'll take it from here.

Adham-A:
For information: I tested on Windows 10

PR #161, opened 2023-10-02
Add a new job parameter: 'background'

### LINKED ISSUE #159: Screen doesn't clear when the file watcher restarts a long-running job
When using the "Run" job with a long-running or never exiting project like a web server, the screen fails to clear when the file watcher triggers a rebuild.

To reproduce:
- Use the "Run" job on a project that doesn't exit when it is run.
- Introduce an error so that the build fails, noting the error in the bacon console.
- Fix the error.
- Note that bacon gets the "computing..." flag, but the output screen continues to show the old error message even though the build was successful and running.
- Force a re-run with the "r" key and note that the screen is now cleared and shows the project output.

I would have expected to see the output of the running project after the file watcher automatically triggers a rebuild.

I was able to fix this issue with the following patch. If this is a fix you'd like I could submit a pull request.

```patch
diff --git a/src/app.rs b/src/app.rs
index 7d10b3d..64e3e7a 100644
--- a/src/app.rs
+++ b/src/app.rs
@@ -104,6 +104,7 @@ pub fn run(
                     debug!("error sending task: {}", e);
                 } else {
                     state.computation_starts();
+                    state.cmd_result = CommandResult::None;
                 }
             }
             recv(executor.line_receiver) -> info => {
```

Canop:
By default true, but should be set to true for never ending jobs

Fix #159


Jayshua:
> By default true, but should be set to true for never ending jobs

I think you meant to write:

> By default true, but should be set to false for never ending jobs

PR #177, opened 2024-03-12
Add option to disable help line

danielwolbach:
Add an option to disable the help line.

This can either be done by using `--no-help-line` or modifying `prefs.toml`:

```
# .config/bacon/prefs.toml
help_line = false
```

The argument `--help-line` overwrites this setting.

The help line doesn't really serve a purpose for me, as it really doesn't show a lot of information and I didn't like it being constantly there.

### DIFF
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -16,6 +16,11 @@
 #
 # reverse = true
 
+# Uncomment and change the value (true/false) to
+# specify whether bacon should show a help line.
+#
+# help_line = false
+
 # Exporting "locations" lets you use them in an external
 # tool, for example as a list of jump locations in an IDE.
 #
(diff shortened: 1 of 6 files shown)


Canop:
Sorry for the delay, I won't forget to evaluate this PR, but it probably  won't be before next week.

Canop:
Now, the question I ask myself is: "Is this option useful enough to justify cluttering the launch options" ?

Other opinions welcome.

danielwolbach:
I would appreciate the cleaner look, as I usually work with a lot of Tmux panes and having less visual clutter always helps. However, I am of course aware that this is purely aesthetic and not necessarily _reasonable_ enough for maintaining additional settings.

Canop:
I merged it. No guarantee it stays but right now there aren't too many options so it's OK.

PR #180, opened 2024-03-30
trigger on_success with warning when allow_warnings

### LINKED ISSUE #179: `on_success` doesn’t trigger when there are warnings and `allow_warnings = true`
Based on the [documentation](https://dystroy.org/bacon/config/#jobs) for `allow_warnings`, I would expect that when `allow_warnings = true`, `on_success` is triggered even when there are warnings:
> if true, the action is considered a success even when there are warnings.

Minimal example (in an empty `cargo` project, current `bacon` `main` 018a2c85a2c223df85df3462caba5eca8da2d969):
```toml
default_job = "check"

[jobs.check]
command = ["cargo", "check", "--color", "always"]
need_stdout = false

[jobs.produce-warning]
command = ["echo", "\u001B[1m\u001B[33mwarning\u001B[0m:", "something"]
need_stdout = true
allow_warnings = true
on_success = "back"

[keybindings]
c = "job:check"
p = "job:produce-warning"
```

I’d like to be able to press `p` while in the `check` job, execute the `produce-warning` job and then automatically go back the the previous job. Is this possible?

If not, *should* this be possible?

My use case is that I want to run `cargo llvm-cov --branch --open` and return to the previous job (similar to `cargo doc --open`). However, `cargo-llvm-cov` unconditionally warns that `--branch` is unstable and there is no way to deactivate that warning.

Canop:
Fix #179

@narpfel Can you please confirm it's OK for you ?

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -440,10 +440,12 @@ impl<'s> AppState<'s> {
     }
     /// the action to execute now
     pub fn action(&self) -> Option<&Action> {
-        self.mission
-            .on_success()
-            .as_ref()
-            .filter(|_| self.cmd_result.is_success())
+        if let CommandResult::Report(report) = &self.cmd_result {
+            if self.mission.is_success(report) {
+                return self.mission.on_success().as_ref();
+            }
+        }
+        None
     }
     fn report_to_draw(&self) -> Option<&Report> {
         self.cmd_result


narpfel:
It works as expected now.

Thanks for the quick fix!

`allow_failures` might need a similar fix? I don’t have a use case for that though.

Canop:
> allow_failures might need a similar fix? I don’t have a use case for that though.

This fixes both

PR #182, opened 2024-04-07
clippy-all job, bound by default to c in bacon.toml

### LINKED ISSUE #167: Difference between how cargo clippy and bacon clippy treat tests
When running ``cargo clippy``, by default, anything behind a ``#[cfg(test)]`` or in the ``tests`` directory will be ignored. ``bacon clippy`` doesn't seem to behave like this - instead, it will run the same lints on the tests too as it always passes the ``--all-targets`` option.

Expected behaviour (at least for me) would be that calling ``bacon clippy`` without further options behaves the same as ``cargo clippy`` does when called without options, and having it run with ``--all-targets`` would either be a custom job or a command line option.

Canop:
- default binding for 'c' in bacon.toml is now the new 'clippy-all' job which does what the old 'clippy' job was doing, ie call clippy on all targets
- 'clippy' job changed to not run on all targets
- default bacon.toml explain how to bind 'c' to clippy instead of 'clippy-all'

Fix #167


twynb:
LGTM :)

PR #183, opened 2024-04-07
expand env vars in job command

### LINKED ISSUE #181: [feature request] Expand env vars in job command
I use env vars as arguments to my executable so I would like support for expanding them before being passed to the executable:

```toml
[jobs.run]
command = ["cargo", "run", "--", "$MY_VAR"]
```

`$MY_VAR` should be expanded before being passed to `cargo run`.

Canop:
Expand env vars in job command unless the job specifies `expand_env_vars = false`.

Fix #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### next
+- expand env vars in job command unless the job specifies `expand_and_vars = false` - Fix #181
+
 <a name="v2.16.0"></a>
 ### v2.16.0 - 2024/03/30
 - `on_success` triggered with warning or errors depending on `allow_warnings` and `allow_failures` - Fix #179
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "bacon"
-version = "2.16.0"
+version = "2.16.1-dev"
 dependencies = [
  "anyhow",
  "cargo_metadata",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bacon"
-version = "2.16.0"
+version = "2.16.1-dev"
 authors = ["dystroy <denys.seguret@gmail.com>"]
 repository = "https://github.com/Canop/bacon"
 description = "background rust compiler"
--- a/bacon.toml
+++ b/bacon.toml
@@ -4,27 +4,28 @@ default_job = "check-all"
 
 [jobs]
 
-[jobs.check-all]
-command = ["cargo", "check", "--all-targets", "--color", "always"]
-need_stdout = false
-watch = ["tests", "benches", "examples"]
-
 [jobs.check]
 command = [
 	"cargo", "check",
 	"--color", "always",
-	"--features", "clipboard",
 ]
 need_stdout = false
 watch = ["benches"]
 
+[jobs.check-all]
+command = [
+	"cargo", "check",
+	"--all-targets",
+	"--color", "always",
+]
+need_stdout = false
+watch = ["tests", "benches", "examples"]
 
 [jobs.win]
 command = [
 	"cross", "build",
 	"--target", "x86_64-pc-windows-gnu",
 ]
-
 
 [jobs.test]
 command = [
--- a/src/job.rs
+++ b/src/job.rs
@@ -7,6 +7,29 @@ use {
 /// One of the possible job that bacon can run
 #[derive(Debug, Clone, Deserialize)]
 pub struct Job {
+    /// Whether to consider that we can have a success
+    /// when we have test failures
+    #[serde(default)]
+    pub allow_failures: bool,
+
+    /// Whether to consider that we can have a success
+    /// when we have warnings. This is especially useful
+    /// for "cargo run" jobs
+    #[serde(default)]
+    pub allow_warnings: bool,
+
+    /// Whether gitignore rules must be applied
+    pub apply_gitignore: Option<bool>,
+
+    /// Whether to wait for the computation to finish before
+    /// to display it on screen
+    ///
+    /// This is true by default. Set it to false if you want
+    /// the previous computation result to be replaced with
+    /// the new one as soon as it starts.
+    #[serde(default = "default_true")]
+    pub background: bool,
+
     /// The tokens making the command to execute (first one
     /// is the executable).
     /// This vector is guaranteed not empty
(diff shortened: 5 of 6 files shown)


twynb (commented):
LGTM (other than the typo) :)

jalil-salame:
I can't seem to get it to work:

```console
$ cargo install --git https://github.com/Canop/bacon --branch expand-env-vars-in-commands --locked
...
    Compiling bacon v2.16.1-dev (/home/jalil/.local/share/cargo/git/checkouts/bacon-73e700fbb85dbb04/5f896ab)
    Finished release [optimized + debuginfo] target(s) in 57.05s
   Replacing /home/jalil/.local/share/cargo/bin/bacon
    Replaced package `bacon v2.16.1-dev (https://github.com/Canop/bacon?branch=expand-env-vars-in-commands#5f896ab8)` with `bacon v2.16.1-dev (https://github.com/Canop/bacon?branch=expand-env-vars-in-commands#5f896ab8)` (executable `bacon`)
$ /home/jalil/.local/share/cargo/bin/bacon --version
bacon 2.16.1-dev
$ cat bacon.toml
...
[jobs.run]
command = [
    "cargo", "run",
    "--color", "always",
    # put launch parameters for your program behind a `--` separator
    "--", "$SUDOKU17"
]
need_stdout = true
allow_warnings = true
background = true
...
$ /home/jalil/.local/share/cargo/bin/bacon -j run
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/sudoku-solver '$SUDOKU17'`
[ERROR]: failed read from file $SUDOKU17: No such file or directory (os error 2)
$ echo $SUDOKU17
/nix/store/pvjqv1ijq416lrjiqpfgbjh4x6h9m5r7-sudoku17
```

Explicitly setting `expand_env_vars = true` doesn't seem to affect it. I even regenerated `bacon.toml` through `bacon --init`.

Am I missing something?

<details><summary><code>bacon.toml</code></summary>

```toml
# This is a configuration file for the bacon tool
#
# Bacon repository: https://github.com/Canop/bacon
# Complete help on configuration: https://dystroy.org/bacon/config/
# You can also check bacon's own bacon.toml file
#  as an example: https://github.com/Canop/bacon/blob/main/bacon.toml

default_job = "check"

[jobs.check]
command = ["cargo", "check", "--color", "always"]
need_stdout = false

[jobs.check-all]
command = ["cargo", "check", "--all-targets", "--color", "always"]
need_stdout = false

[jobs.clippy]
command = [
    "cargo", "clippy",
    "--all-targets",
    "--color", "always",
]
need_stdout = false

# This job lets you run
# - all tests: bacon test
# - a specific test: bacon test -- config::test_default_files
# - the tests of a package: bacon test -- -- -p config
[jobs.test]
command = [
    "cargo", "test", "--color", "always",
    "--", "--color", "always", # see https://github.com/Canop/bacon/issues/124
]
need_stdout = true

[jobs.doc]
command = ["cargo", "doc", "--color", "always", "--no-deps"]
need_stdout = false

# If the doc compiles, then it opens in your browser and bacon switches
# to the previous job
[jobs.doc-open]
command = ["cargo", "doc", "--color", "always", "--no-deps", "--open"]
need_stdout = false
on_success = "back" # so that we don't open the browser at each change

# You can run your application and have the result displayed in bacon,
# *if* it makes sense for this crate.
# Don't forget the `--color always` part or the errors won't be
# properly parsed.
# If your program never stops (eg a server), you may set `background`
# to false to have the cargo run output immediately displayed instead
# of waiting for program's end.
[jobs.run]
command = [
    "cargo", "run",
    "--color", "always",
    # put launch parameters for your program behind a `--` separator
    "--", "$SUDOKU17"
]
expand_env_vars = true
need_stdout = true
allow_warnings = true
background = true

# This parameterized job runs the example of your choice, as soon
# as the code compiles.
# Call it as
#    bacon ex -- my-example
[jobs.ex]
command = ["cargo", "run", "--color", "always", "--example"]
need_stdout = true
allow_warnings = true

# You may define here keybindings that would be specific to
# a project, for example a shortcut to launch a specific job.
# Shortcuts to internal functions (scrolling, toggling, etc.)
# should go in your personal global prefs.toml file instead.
[keybindings]
# alt-m = "job:my-job"
```

</details>

Canop:
OK... that was a stupid bug...

    r"\$([A-Z_]+)"

Canop:
@jalil-salame Can you try again ?

jalil-salame:
> Note: It now also accepts lowercase letters. This is theoretically legal but rarely used. So I'm not sure about whether it's a good idea.

Lowercase env vars are generally for local variables (in shell scripts) but notably, when using proxies (e.g. `HTTP_PROXY=xxx`) tools like `curl` accept both uppercase and lowercase version of the variables (e.g. `HTTP_PROXY` and `http_proxy` are accepted).

According to <https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08>:

> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2017 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names. Uppercase and lowercase letters shall retain their unique identities and shall not be folded together. The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.

I would assume most people would limit themselves to `[A-Za-z0-9_]`, if I take a look at my env, all variables start with `[A-Za-z_]`, they also only contain only `[A-Za-z0-9_]`.

I don't think a regex is needed, [`strip_prefix('$')`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) and [`str.bytes().all(|c| c == b'_' || c.is_ascii_alphanumeric())`](https://doc.rust-lang.org/std/primitive.u8.html#method.is_ascii_alphanumeric) is enough.

jalil-salame:
> @jalil-salame Can you try again ?

It works perfectly now.

Canop:
> I don't think a regex is needed, [strip_prefix('$')](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) and [str.bytes().all(|c| c == b'_' || c.is_ascii_alphanumeric())](https://doc.rust-lang.org/std/primitive.u8.html#method.is_ascii_alphanumeric) is enough.

With the` added expand-env-vars-in-commands` setting, I decided to support env vars in internal positions, eg

```
command = [
    "cargo", "run",
    "--color", "always",
    "--", "$PREFIX-$SUFFIX",
]
```

jalil-salame:
> With the` added expand-env-vars-in-commands` setting, I decided to support env vars in internal positions, eg

I didn't notice that, that's cool! Yeah, regex makes more sense there. If we already had a dependency on nom/winnow we could do it with that instead and it _should_ be faster, but no sense in adding more dependencies.

PR #184, opened 2024-04-14
Parse test results from tests running with `--quiet`

narpfel:
Running tests with `-q`/`--quiet` outputs a `.` per test instead of the test name. When a test fails, the test name and `FAILED` is still printed, but the `test ` at the beginning of the line is omitted:
```
# run without `--quiet`
test tests::failing_test ... FAILED
# run with `--quiet`
tests::failing_test --- FAILED
```
This PR changes the regex used to parse the output to parse both output formats.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - default binding for 'c' in bacon.toml is now the new 'clippy-all' job which does what the old 'clippy' job was doing. 'clippy' job changed to not run on all targets. Default bacon.toml explain how to bind 'c' to clippy instead of 'clippy-all' - Fix #167
 - expand env vars in job command unless the job specifies `expand_env_vars = false` - Fix #181
 - some file events filtered out from watch (feedback welcome, especially if you notice some failures to recompute)
+- parse test results even when tests are run with `-q`/`--quiet`
 
 <a name="v2.16.0"></a>
 ### v2.16.0 - 2024/03/30
--- a/src/line_analysis.rs
+++ b/src/line_analysis.rs
@@ -130,7 +130,7 @@ fn is_build_failed(ts: Option<&TString>) -> bool {
 /// part is in another TString)
 fn as_test_name(s: &str) -> Option<&str> {
     regex_captures!(
-        r#"^test\s+(.+?)(?: - should panic\s*)?(?: - compile\s*)?\s+...\s*$"#,
+        r#"^(?:test\s+)?(.+?)(?: - should panic\s*)?(?: - compile\s*)?\s+...\s*$"#,
         s
     )
     .map(|(_, key)| key)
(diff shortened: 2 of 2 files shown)


Canop:
Can you please build a test case which failed and is solved ?

narpfel:
Here’s an example that shows this:

`bacon.toml`:
```toml
default_job = "test"

[jobs.test]
command = [
    "cargo", "test", "--color", "always",
    "--", "--color", "always", # see https://github.com/Canop/bacon/issues/124
]
need_stdout = true

[jobs.testquiet]
command = [
    "cargo", "test", "--color", "always",
    "--", "--color", "always", # see https://github.com/Canop/bacon/issues/124
    "--quiet",
]
need_stdout = true


[keybindings]
z = "job:testquiet"
```

`src/lib.rs`:
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn first_test() {}

    #[test]
    fn failing_test() {
        assert!(false, "message for first failing test");
    }

    #[test]
    fn third_test() {}

    #[test]
    fn another_failing_test() {
        assert!(false, "message for second failing test");
    }
}
```

Output when running the `test` job:
```
 t   test    2 fails

 1  failed: tests::failing_test
 thread 'tests::failing_test' panicked at src/lib.rs:8:9:
 message for first failing test

 2  failed: tests::another_failing_test
 thread 'tests::another_failing_test' panicked at src/lib.rs:16:9:
 message for second failing test
```

Output when running the `testquiet` job with the `bacon` 2af07e74c03de51b709b3a8e41dad059583a65e6 (current `main`):
```
 t   testquiet   Command error code: 101

    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (target/debug/deps/t-32496683b92ff7c2)

running 4 tests
tests::another_failing_test --- FAILED
. 2/4
tests::failing_test --- FAILED
.
failures:

---- tests::another_failing_test stdout ----
thread 'tests::another_failing_test' panicked at src/lib.rs:16:9:
message for second failing test
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::failing_test stdout ----
thread 'tests::failing_test' panicked at src/lib.rs:8:9:
message for first failing test


failures:
    tests::another_failing_test
    tests::failing_test

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass `--lib`
```

And with this PR:
```
 t   testquiet    2 fails

 1  failed: tests::failing_test
 thread 'tests::failing_test' panicked at src/lib.rs:8:9:
 message for first failing test

 2  failed: tests::another_failing_test
 thread 'tests::another_failing_test' panicked at src/lib.rs:16:9:
 message for second failing test
```

Or did you mean an automated test?

Canop:
> Or did you mean an automated test?

No, just enough to reproduce the problem.

Canop:
Here's why I asked: it doesn't seem to be solved with your fix (in my test on linux).

I'll look deeper if necessary but not immediately, I can't dive into this problem until end of last week.

narpfel:
> it doesn't seem to be solved with your fix (in my test on linux, even with your test case).

I’m on `nightly`, and hadn’t realised that the output format was recently changed: https://github.com/rust-lang/rust/pull/113518 So on `stable` (currently 1.77) this PR will not work (because the test name and `FAILED` are not printed), but on Rust 1.78+ it will work.

Sorry for the confusion!

PR #187, opened 2024-05-18
feat: allow context to be stored in location export

crisidev:
I have created a LSP server based on Bacon (great software btw) called [bacon-ls](https://github.com/crisidev/bacon-ls) ([blog post](https://lmno.lol/crisidev/bacon-language-server)) that makes use of the location export.

This PR allows a new setting called `add_context_to_message` changing the line export to include all the context from the compiler.

### DIFF
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -29,8 +29,15 @@
 # Set 'enabled' to true to have bacon always export locations
 # This is equivalent to always adding -e to bacon commands
 # but can still be cancelled on specific launches with -E
+#
+# 'add_context_to_message' is used to control whether to ad
+# any normal line linked to the diagnostic title. When this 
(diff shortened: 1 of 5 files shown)


Canop:
Rather than a `add_context_to_message` setting, I'd prefer to allow changing the `line_format` to

`line_format = "{kind} {path}:{line}:{column} {message}{context}"`

The purpose of the `line_format` parameter is to allow selecting and partially format what's in the exported line.

This doesn't prevent optimization: we can check whether the format contains `{context}` before calling `extract_diagnostic_context`.

If you prefer I do the change, I can do it.

crisidev:
This is fine by me, it achieves the same result and it's a better approach. If you want to implement the change, be my guest :)

Canop:
Follow up: https://github.com/Canop/bacon/pull/188

crisidev:
Superseded by #188

PR #192, opened 2024-05-30
fix failure to recognize location in test compilation output

### LINKED ISSUE #190: Show the error line number in doc test summaries instead of the block line number.
When doc tests fail, the cargo test output contains the line number of the start of the block like:
`test src/buffer/buffer.rs - buffer::buffer::Buffer::index (line 465) ... FAILED`
and then the actual error later:
```
---- src/buffer/buffer.rs - buffer::buffer::Buffer::index (line 465) stdout ----
error[E0507]: cannot move out of index of `ratatui::buffer::Buffer`
 --> src/buffer/buffer.rs:469:12
```
The line with the src/buffer/buffer.rs:469:12 is the location of the error. Showing this location in the summary rather than the line number of the start of the block would make this clickable in the VSCode terminal - this makes it easy to navigate to broken tests and fix them fast. Doing so from the summary rather than the detail view is especially appealing for longer running tests like the doc tests are generally.

Canop:
Fix #190

### DIFF
--- a/src/line_analysis.rs
+++ b/src/line_analysis.rs
@@ -37,6 +37,9 @@ impl From<&CommandOutputLine> for LineAnalysis {
             } else if regex_is_match!(r#", [^:\s'"]+:\d+:\d+$"#, content) {
                 // this kind of location comes up in test failures
                 LineType::Location
+            } else if regex_is_match!(r#"^\s+--> [^:\s'"]+:\d+:\d+$"#, content) {
+                // this comes up in test failures to compile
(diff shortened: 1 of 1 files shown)


joshka:
Tested against the original error - looks good to me. I especially like how this displays multiple errors as two lines in the output.

Canop:
I suspect the reason I had to change this is a regression in `cargo test` which doesn't seem to respect the `--color always` argument anymore.

PR #198, opened 2024-07-05
Fix a typo in config.md

DenisGorbachev:


Canop:
This was bad English but not a typo. I guess a better word would be "quickly", or "soon".

DenisGorbachev:
Done

PR #199, opened 2024-07-22
added option `extraneous_args`

TheTollingBell:
I was annoyed at the inability to integrate `cargo fmt` as a bacon task easily alongside `--all-features` so I added an option `extraneous_args` which enables and disables functionality for modifying arguments in a `Job` during its conversion to a `Command`.


Canop:
You mean that what you needed was a way for a job to ignore the arguments given at launch ?

TheTollingBell:
Yes, I just wanted a way to run non-cargo tasks every once and a while on a bind.

PR #202, opened 2024-08-09
Add kill command to job

pcapriotti:
First, thanks for this awesome project!

This PR adds an optional job field `kill`, which specifies a custom command to be used for killing a job. It can be used to make sure that commands are given a chance to clean up when they are terminated by bacon.

If the `Command` API offered a better way to kill child processes, maybe killing them with SIGTERM could be the default. But as far as I can tell there is no easy portable way to do that without depending on some other crate. Letting the user specify a command for those cases where it is important that processes are killed cleanly seems a reasonable compromise.

### DIFF
--- a/defaults/default-bacon.toml
+++ b/defaults/default-bacon.toml
@@ -82,6 +82,7 @@ command = [
 need_stdout = true
 allow_warnings = true
 background = true
+# kill = ["kill", "-s", "TERM"]
 
 # This parameterized job runs the example of your choice, as soon
 # as the code compiles.
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -14,6 +14,7 @@ use {
             BufReader,
         },
         process::{
+            Child,
             Command,
             Stdio,
         },
@@ -27,6 +28,7 @@ use {
 /// Channel sizes are designed to avoid useless computations.
 pub struct MissionExecutor {
     command: Command,
+    kill_command: Vec<String>,
     /// whether it's necessary to transmit stdout lines
     with_stdout: bool,
     line_sender: Sender<CommandExecInfo>,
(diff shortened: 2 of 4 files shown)


Canop:
Can you give me a concrete example where it would solve a problem and the kill command you would use in that case ?

pcapriotti:
Sure.

Say you're working on a long-running service like a web application, and you want to run it in bacon using `cargo run`. If the service allocates some resources that need to be disposed of on shutdown (e.g. it creates temporary files, or spawns child processes of its own), killing it abruptly at every refresh leaves those resources around, and might cause problems on the next run, or just cause garbage to pile up over time.

So in cases like these, you would use something like
```toml
kill = ["kill", "-s", "INT"]
```
to send a SIGINT to the process and give it a chance to terminate cleanly.

Canop:
I don't know what tests you did until now. Can I assume you tried your kill feature for some time, with many kills, and it worked well ?

pcapriotti:
Yes, I tested it manually. For example, given the following `main.rs`:
```rust
use std::sync::mpsc;

fn main() {
    let (tx, rx) = mpsc::channel();

    {
        let tx = tx.clone();
        ctrlc::set_handler(move || {
            println!("interrupted!");
            tx.send(()).unwrap();
        })
        .unwrap();
    }

    let file = tempfile::NamedTempFile::new().unwrap();
    println!("{}", file.path().display());

    std::thread::spawn(move || {
        std::thread::sleep(std::time::Duration::from_secs(10));
        tx.send(()).unwrap();
    });

    rx.recv().unwrap();
}
```
Then run bacon, press 'r' and refresh it a few times. After quitting, all the temporary files are left lingering. Now add 
```
kill = ["kill", "-s", "INT"]
```
to `bacon.toml` and the temporary files will be deleted as expected on every refresh.

Canop:
I added some documentation and merged it. Thanks, I think this is a useful addition.

PR #203, opened 2024-08-11
Issue 157

### LINKED ISSUE #157: Default `bacon.toml` overrides `prefs.toml` configured jobs
I have some default jobs to use on projects that haven't configured `bacon` themselves in my `prefs.toml`. To use them I still need to `touch bacon.toml` in each project to stop the default package config overriding my system configuration. It feels like these defaults should not override anything that has been explicitly configured in `prefs.toml`.

Canop:
Apply the default bacon.toml before the prefs.toml and the possible user defined bacon.toml files.

This allows overriding the default in prefs.toml.

Fix #157

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -67,6 +67,7 @@ command = [
 	"-A", "clippy::len_without_is_empty",
 	"-A", "clippy::get_first",
 	"-A", "clippy::while_let_on_iterator",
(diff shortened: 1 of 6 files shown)


Canop:
This will be a "breaking change", as the end result of some configurations won't be the same.

But it's probably not worth changing the major of bacon.

PR #207, opened 2024-09-07
don't kill jobs on every change

### LINKED ISSUE #206: Feature Suggestion: Show number of files changed since last run in header line
I often make changes to files while bacon is still compiling (esp. before the pause functionality was added). This usually results in bacon showing slightly outdated information (which is okay), but unfortunately without any hint that it is outdated.

Ideally bacon would show a `files-changed=5` hint in the first row if the file system watcher detects that a file changed while bacon is still compiling (and thus bacon not re-starting the compilation). This should be updated immediately and should be reset to 0 when the job / a new compilation starts which includes the changed files.

I'm not sure if/when exactly cargo detects/includes these files into the compilation, but such a hint in bacon would at least indicate that the output might not include all changed files. It would also be useful when bacon is paused, to indicate how many files you've changed since bacon was paused (that is of course assuming that the filesystem watcher is still running while compiling and while bacon is paused).

Canop:
This PR brings 2 settings:

With `show_changes_count = true`, you can see the number of file changes that occurred since last job start.

The `on_change_strategy` setting lets you define how bacon reacts to a change, with two possible values:

* With `on_change_strategy = "kill_then_restart"`, the current job is immediately killed and a new job restarted. This is the behavior that bacon had before this PR.  It has the downside of never allowing any job to complete if you're always changing files and the job is just a little too long to finish between changes.

* With `on_change_strategy = "wait_then_restart"` (which is the new default, so you can omit it), bacon waits for the job to finish before restarting it. This is probably much better when the jobs aren't instant and you want to continue changing files while it's computing.

The `on_change_strategy` can be defined in the global prefs, in the project settings, and even for a specific job.

Fix #206

### DIFF
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -21,6 +21,10 @@
 #
 # help_line = false
 
+# Uncomment and change the value (true/false) to
+# set whether to display the count of changes since last job start
+# show_changes_count = false
+
 # Exporting "locations" lets you use them in an external
 # tool, for example as a list of jump locations in an IDE
 # or in a language server.
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -23,7 +23,15 @@
 
 # Uncomment and change the value (true/false) to
 # set whether to display the count of changes since last job start
+#
 # show_changes_count = false
+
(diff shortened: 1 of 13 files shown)


Canop:
released in 2.21.0

PR #219, opened 2024-10-01
toggle-backtrace accepts an optional level: 1 or full

### LINKED ISSUE #210: `RUST_BACKTRACE` is silently ignored
I tried to activate backtraces by setting `env.RUST_BACKTRACE = "1"` and was confused that no backtraces were shown. It didn’t occur to me to check the docs because (for me) it was very obvious that setting `RUST_BACKTRACE=1` should enable backtraces, so I thought that something was wrong with my program that prevented backtraces from being shown.

It would have helped me if bacon had warned that setting `RUST_BACKTRACE` has no effect.

Also, from skimming the code, apparently it is supposed to show `press b to toggle backtraces` in the status line when there is a panic message in the output? That doesn’t happen for me with the default `run` profile (no `bacon.toml` file) (pressing `b` works though, it just is a bit undiscoverable):

```rust
fn main() {
    panic!()
}
```
=>
```console
$ bacon run
 t   run   Command error code: 101

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/t`
thread 'main' panicked at src/main.rs:2:5:
explicit panic
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Hit q to quit, w to not wrap lines, p to pause, ? for help
```

Am I missing something?

Also, another small thing: When using crates like [`color_eyre`](https://docs.rs/color-eyre/latest/color_eyre/), it would be nice to be able to set `RUST_BACKTRACE=full` because then the backtraces show the source for each stack frame (there’s an example of that [here](https://docs.rs/color-eyre/latest/color_eyre/#multiple-report-format-verbosity-levels)).

Canop:
Fix #210

With this PR, it's possible to decide that the backtrace level, when enabled, won't be "1" but "full" by adding this to your bacon.toml

```
[keybindings]
b = "toggle-backtrace(full)"

```

Of course you can bind to a different key.

This is experimental and may change depending on feedback (if you use it, do report).


narpfel:
Nice, thanks! This looks great.

While testing, I noticed two things that could maybe be improved:
* Switching between different backtrace levels would be nice. I tried to add `shift-b = "toggle-backtrace(full)"` to my config in addition to the default `b`. This leads to the following: no backtrace → press `b` → level 1 → press `shift-b` → no backtrace. It would be nice if you could switch directly from level 1 to level full.
* `color_eyre` modifies the panic message to this:

  ```
  Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
  Run with RUST_BACKTRACE=full to include source snippets.
  ```

  Maybe the regex here https://github.com/Canop/bacon/blob/c4be3fcd77358eedbde30f75ee90c1a8ed9b1180/src/line_analysis.rs#L39 could be relaxed a bit, e. g. to ``[Rr]un with `RUST_BACKTRACE=``? Just recognising the default is already great though.

Canop:
You're right. Done and done.

PR #231, opened 2024-10-07
Fix prefs.toml path quoting

teor2345:
On macOS, when I run:
`$EDITOR $(bacon --prefs)`
My editor tries to open:
`/Users/name/repository/Support/org.dystroy.bacon/prefs.toml`,
rather than the correct path:
`/Users/name/Library/Application Support/org.dystroy.bacon/prefs.toml`.

This can be fixed by quoting the path, either in `bacon`, or in the documentation. This PR fixes `bacon`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,6 +142,7 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_json",
+ "shlex",
  "termimad",
  "toml",
  "unicode-width",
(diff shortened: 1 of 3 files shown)


Canop:
I think this fix is better: https://github.com/Canop/bacon/commit/bde8a0d6a321bb0ed38cb8d5ff9771db519bd475

I'm sorry it's not yet fixed on the web site: I should deploy the new site when I deploy the new version of bacon, which would happen soon if no bug is found.

PR #233, opened 2024-10-07
ignore events during grace period after task start

### LINKED ISSUE #222: vim double-triggers bacon v2.21.0 when saving a file
When saving a file in vim (or neovim), vim will write to multiple files in the directory of the saved file. Here is a log for saving `src/main.rs` in neovim:

```console
$ inotifywait --monitor src
Setting up watches.
Watches established.
src/ CREATE 4913
src/ OPEN 4913
src/ ATTRIB 4913
src/ CLOSE_WRITE,CLOSE 4913
src/ DELETE 4913
src/ MOVED_FROM main.rs
src/ MOVED_TO main.rs~
src/ CREATE main.rs
src/ OPEN main.rs
src/ MODIFY main.rs
src/ CLOSE_WRITE,CLOSE main.rs
src/ ATTRIB main.rs
src/ DELETE main.rs~
```

(Note that this doesn’t happen on `tmpfs` filesystems, there neovim directly writes the file. I can reproduce the shown behaviour on `btrfs` and `ext4`; I don’t have any other filesystems at hand.)

This means that with the new default for `on_change_strategy` introduced in #207, each save will run the current job (e. g. running the tests) twice. The old behaviour works well, so feel free to ignore this issue, but it might be nice to avoid retriggering anything in this situation (maybe by adding a timeout of a couple milliseconds?).

Canop:
When some file events are received:

* the task child process waits for 5ms before starting the real task
* during the same period, file events are discarded

The goal is to prevent
* the cargo job from running with not all changes
* a bunch of quasi-simultaneous file changes from triggerring more than one cargo execution

Downside is that you have to wait for 5ms more before the computation result is displayed after a save.

The delay can be removed or modified in the preferences. Eg `grace_period="15ms"` or `grace_period="none"`.

May fix #222


narpfel:
Thanks, this works well for me.

> * the task child process waits for 5ms before starting the real task

I’m not sure if that’s really necessary. Yes, there technically is a race between the text editor writing the file and the triggered job reading the file, however I’d expect that in the vast majority of cases, the text editor is faster. Just starting `cargo` through the `rustup` shim takes ~50ms (https://github.com/rust-lang/rustup/issues/2626), so the sleep here https://github.com/Canop/bacon/blob/b8cb63627e82cf523ec62455714a21565cff29bf/src/exec/executor.rs#L111-L116 is probably not necessary. On the other hand, a couple of milliseconds are probably not noticeable.

Canop:
I first pushed the PR without this sleep, then I decided I didn't want to have race conditions when cargo becomes faster and I'm sure nobody will be annoyed by those 5 or 10 ms.

PR #235, opened 2024-10-07
fix either raw/standard output not rewrapped

### LINKED ISSUE #234: Trying to wrap or unwrap lines in raw mode does nothing
On macOS, when I use `toggle-raw-output` to go back to the original verbose test output, I can't wrap or unwrap lines.

Steps to reproduce:
1. Run a test with long lines and failures
2. Wait for it to finish, which goes to the fails summary screen
3. Do `toggle-raw-output` (I use my own key binding for this)
4. Try to wrap or unwrap lines

Expected outcome:
Line wrapping changes

Actual outcome:
Line wrapping does not change

If raw mode ignored wrapping entirely, that would be consistent but annoying. Instead, raw mode won't unwrap-to-wrap or wrap-to-unwrap lines.

Canop:
Fix #234

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -165,6 +165,10 @@ impl<'s> AppState<'s> {
     }
     pub fn toggle_raw_output(&mut self) {
         self.raw_output ^= true;
+        if self.wrapped_output.is_some() {
+            self.wrapped_output = None;
+            self.update_wrap(self.width - 1);
+        }
     }
     pub fn set_result(
         &mut self,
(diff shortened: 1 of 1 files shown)


teor2345:
Thanks for the quick fix, but this doesn't seem to quite work the way I expect.

When I set `wrap = false` in `prefs.toml`, no wrapping happens at all, either in the initial `test` screen, or the `fails` result summary screen, regardless of how many times I press `w`. I expected that `w` would cause wrapping (and then unwrapping when pressed again), in both screens.

When I set `wrap = true` in `prefs.toml`, wrapping happens in the `fails` screen, and correctly unwraps on `w`. This works as expected. But the raw output starts unwrapped, and doesn't wrap on `w`.

Is there another way of switching back to the detailed output that also does wrapping? Because maybe I'm expecting the wrong things from raw mode. If that's the case, then the only bug is disabling wrapping entirely in the `fails` screen when `wrap = false`.

Canop:
You're right. I was too fast in pushing this PR.

Canop:
I think I fixed the fix.

teor2345:
Sory for the delay, yes this works as expected!

Issue #216, opened 2024-10-01
Relaunch blocking tasks


musjj:
It looks like that blocking tasks (e.g. running a web server) aren't automatically restarted when there are changes. Is there a way to configure so that it's automatically killed and restarted?

musjj:
Oh it looks like that there was a change in the last version:

https://github.com/Canop/bacon/releases/tag/v2.21.0

> With `on_change_strategy = "kill_then_restart"`, the current job is immediately killed and a new job restarted. This is the behavior that bacon had before this PR. It has the downside of never allowing any job to complete if you're always changing files and the job is just a little too long to finish between changes.

Canop:
Yes. I reopen the issue because I should improve the default bacon.toml file and maybe the doc to explain that

Canop:
website updated: https://dystroy.org/bacon/config/#jobs

But I'll add a comment in the default bacon.toml too

PR #245, opened 2024-10-18
Add note about running servers

kdwarn:
I racked my brain for a little while about how to get my Axum server to rerun after changes were made while running bacon, and discovered the answer in https://github.com/Canop/bacon/issues/216, which pointed to updates made to the default bacon.toml file. The place I initially looked to try to solve this was in the [Cookbook](https://dystroy.org/bacon/cookbook/), under the section on running binaries. So, to make discovering how to resolve this issue easier for others, I took essentially what you wrote in the default bacon.toml file and added it to the Cookbook.

### DIFF
--- a/website/docs/cookbook.md
+++ b/website/docs/cookbook.md
@@ -97,6 +97,17 @@ Depending on the desired output, you would have to add a setting to the run job,
 command = ["cargo", "run", "--color", "always", "--", "--color", "yes"]
 ```
 
+## Servers
+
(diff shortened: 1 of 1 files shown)


Canop:
This part, and the "Gentler Kill" one just after, should probably be merged into a "Settings for long running programs" chapter

kdwarn:
I think maybe I’d rather let you do it - probably would end up with less work for you in the end than going back and forth with it.

Canop:
Done: https://dystroy.org/bacon/cookbook/#long-running-programs

If you think it can be improved, or that it wouldn't have helped you find the right config, please tell

kdwarn:
Looks good to me!

PR #247, opened 2024-10-20
Update default-bacon.toml to add run-long job

### LINKED ISSUE #246: [Feature] Make long-running jobs easier to work with
I've been using cargo-watch to re-compile my web server as I iterate on it. I simply run `cargo watch -w src/ -x run` in my terminal.

With Bacon, I must first add a `bacon.toml` file to the root of my project directory and then modify the run job in that file to the following:
```toml
[jobs.run]
command = [
    "cargo", "run",
    "--color", "always",
]
need_stdout = true
allow_warnings = true
background = false # Needed for long-running jobs!
on_change_strategy = "kill_then_restart" # Needed for long-running jobs!
```

It took me a good amount of time to figure out this configuration was needed. I wonder if we can make long-running jobs easier to work with. I would guess that working with servers is a common use case. Ideally, no config file would be necessary, so end-users can simply run a command in the terminal. 

Here are some options I have in mind:
1. Create a default job called “run-long” which has the `background` and `on_change_strategy` job fields set appropriately. 
    - Pros
        - Simple and easy to discover through the CLI --help menu
        - Hides job config details from end-users
        - Very easy to implement
    - Cons
        - Too niche? Not sure if this exact config is popular enough to warrant its own default job. 
        - Currently, jobs are differentiated by command arguments, not job fields when you run `bacon --list-jobs`. This means the `run-long` job would look exactly the same as the `run` job. We would need to add some information noting the difference between these two jobs.
2. Allow job fields to be specified via CLI arguments
    1. Cargo-watch example: 
        1. `cargo watch -q -c -w src/ -x run`
    2. Bacon examples: 
        1. `bacon run —background false —on-change-strategy kill-then-restart`
        2. `bacon run —bg false —on-change kill`
    3. Pros/Cons   
        - Pros
            - This solution allows for fine-grain configuration through CLI arguments and could be useful outside of the long-running jobs use case.
        - Cons
           - Scope creep? Not sure if exposing job fields is generally useful. If not, then doing this only for this one use case might be overkill.  
3. Do both options 1 and 2

Thoughts?

ryancannon88:
This change resolves #246 by adding a new `run-long` job to `default-bacon.toml`.

### DIFF
--- a/defaults/default-bacon.toml
+++ b/defaults/default-bacon.toml
@@ -75,10 +75,6 @@ on_success = "back" # so that we don't open the browser at each change
 # *if* it makes sense for this crate.
 # Don't forget the `--color always` part or the errors won't be
 # properly parsed.
-# If your program never stops (eg a server), you may set `background`
-# to false to have the cargo run output immediately displayed instead
-# of waiting for program's end. If you prefer to have it restarted at
-# every change, then uncomment the 'on_change_strategy' line.
 [jobs.run]
 command = [
     "cargo", "run",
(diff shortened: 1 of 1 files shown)


Canop:
merged. Thanks.

PR #248, opened 2024-10-25
allow overriding default watched directories/files

nolanderc:
Adds a new global configuration option `default_watch` to override the default set of files/directories used to detect changes (e.g. `["src", "tests", "benches", "examples", "build.rs"]`).

### DIFF
--- a/defaults/default-bacon.toml
+++ b/defaults/default-bacon.toml
@@ -6,6 +6,7 @@
 #   https://github.com/Canop/bacon/blob/main/defaults/default-bacon.toml
 
 default_job = "check"
+default_watch = ["src", "tests", "benches", "examples", "build.rs"]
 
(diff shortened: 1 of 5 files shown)


Canop:
This makes sense but the name isn't right.

Some settings can be found either in a job or at the root of a configuration file, for example `on_change_strategy`. The one at the root is applied to all jobs unless overridden in a job. This logic will probably be applied to a few other settings.

This new setting should be called `watch`, not `default_watch`.
BTW another setting can become necessary: `default_watch` with a bool value, as is already present in the job.

Canop:
Due to the consistent naming problem I mentioned, I finally did https://github.com/Canop/bacon/pull/253

Thanks a lot for the PR and valid idea, though

PR #252, opened 2024-10-28
docs: Add details for how to set default job.

Jasha10:
Updated `config.md`


Canop:
Do you propose this because it would have helped you ?

Jasha10:
Yes, it would have helped me :)

PR #256, opened 2024-11-08
allow bacon to run without Cargo.toml file

Canop:
This makes it possible to launch bacon in non rust repositories to run non rust-related jobs.


Canop:
Tests on complex rust projects or setups would be welcome, to check I didn't mess anything.

PR #260, opened 2024-11-12
'python_unittest' analyzer

### LINKED ISSUE #259: Python unit test analyzer
Make it possible to test python with bacon

Canop:
Fix #259

It seems to work.

Summarized:
image

Normal:
image

Jumping to locations in nvim works too.

To obtain this, you need a bacon.toml file containing for example

```TOML
default_job = "test"

[jobs.test]
command = [
    "python3", "src/truc.py",
]
need_stdout = true
watch = ["src"]
analyzer = "python_unittest"

```

Now, let's be real: I didn't write any serious Python code in the last 15 years, so I need some adults to come and tell me what doesn't really work and what should be supported.

### DIFF
--- a/src/analysis/analyzer.rs
+++ b/src/analysis/analyzer.rs
@@ -2,6 +2,7 @@ use {
     super::{
         eslint,
         nextest,
+        python,
         standard,
     },
     crate::*,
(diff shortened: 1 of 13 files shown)


Canop:
I might do pytest too.

Issue #261, opened 2024-11-16
`--color always` revisited


joshka:
Cargo has a CARGO_TERM_COLOR environment variable which can be used to force colored output.
Adding a new top level item that allows environment variables to be set for all commands seems like it might solve https://github.com/Canop/bacon/issues/124 properly without having to litter commands with `--color always` in the config.

I'd suggest expanding https://github.com/Canop/bacon/pull/146 to make global environment vars.

The workaround of setting these in the .cargo/config.toml is not quite right as that would affect all cargo usages, but I only want to affect usages of cargo from Bacon.

Canop:
This is very interesting, I didn't know this variable.

joshka:
> This is very interesting, I didn't know this variable.

Neither did I until yesterday :)

Issue #124, opened 2023-03-25
bacon test doesn't color tests


svenstaro:
Hey, cool project! This is kind of a minor issue but I still think it'd be nice to have tests be able to output in color. However, `bacon test` doesn't use color unless forced to with `bacon test -- -- --color always` which is kind of inconvenient.

There is this related issue: https://github.com/rust-lang/cargo/issues/1983

Canop:
Can you build a test case exhibiting the lack of color ?

svenstaro:
Result of `bacon test` on a random Rust project with tests:

![image](https://user-images.githubusercontent.com/1664/227705644-229c6ba6-7181-4159-ad6e-c2c5c2c71ffc.png)

Running it with `bacon test -- -- --color always`:
![image](https://user-images.githubusercontent.com/1664/227705690-74ab9690-332b-4ff6-a856-e3b5de569695.png)

Canop:
Ok, I see.

So this could be fixed with this definition of the test job (wich I could make default):

```
[jobs.test]
command = ["cargo", "test", "--color", "always", "--", "--color", "always"]
need_stdout = true
```

Any better suggestion ?

svenstaro:
I think that'd work. Since bacon is only ever meant to run interactively, I don't see any issues here.

PR #263, opened 2024-11-16
Add root env vars to fix --color=always problem

joshka:
Adds a new root level config item env, which allows setting environment
vars for all jobs. This makes it possible then to set CARGO_TERM_COLOR
to always, which fixes the problem with --color=always not working.

Fixes: #261 and #124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - more lenient detection of warnings and errors due to 'miri run' not supporting `--color` - Fix #251
 - eslint analyzer (set `analyzer = "eslint"` in your job definition)
 - new `ignore` job parameter, accepts a list of glob patterns
+- allow defining environment vars for all jobs - Fix #261 and #124
 
 <a name="v3.2.0"></a>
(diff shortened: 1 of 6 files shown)


Canop:
Something which may seem a little disturbing: if you remove the `env.CARGO_TERM_COLOR = "always"` line from the bacon.toml of your project, the variable is still set, because the default bacon.toml is used to init settings and env vars are only extended.

This doesn't look like it can be a problem, it's even good as it will make it possible for users to add new jobs without the `--color always` thing even if they don't have the `env.color = always` line at the top.

But the fact that you can't, today, remove an env var defined previously (in the prefs.toml or in the bacon.toml root) may be seen as a problem. Maybe we'll have to add an ugly `inherit_env = false` in the future ?

joshka:
Yeah, I was thinking about that as a potential problem too. Perhaps making the hashmap accept `Option<String>` would work for unsetting vars.

I also didn't add env to the args, which might be nice too, mainly because I wasn't sure the syntax for representing `--env FOO=bar` in clap (and didn't find an obvious answer in a few minutes of search)

PR #271, opened 2024-12-02
Biome analyzer

Canop:
For use with https://biomejs.dev/

Example config:

```
[jobs.biome-libs]
env.RAYON_NUM_THREADS = "1" # for constant ordering of items
command = [
    "npx", "@biomejs/biome", "lint",
    "--colors", "force",
    "./libs",
    "--skip", "complexity/useArrowFunction",
    "--skip", "style/useTemplate",
]
need_stdout = true
analyzer = "biome"
watch = ["libs"]
```

Summarized:

image

Not summarized:

image

The env thing gives a reproducible order of items.

### DIFF
--- a/website/docs/config.md
+++ b/website/docs/config.md
@@ -111,10 +111,11 @@ Available analyzers:
 * `standard` (default)
 * `nextest`
 * `eslint` for [ESLint](https://eslint.org/)
+* `biome` for [Biome](https://biomejs.dev/)
 * `python_unittest` for [Unittest](https://docs.python.org/3/library/unittest.html)
 * `python_pytest` for [pytest](https://docs.pytest.org/)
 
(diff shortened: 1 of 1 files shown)


src/analysis/biome/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+                    code,
+                    tag,
+                });
+            }
+        }
+    }
+    None
+}
+
+/// Build a report from the output of eslint
```
tyilo, line 89: Mention biome instead?

Canop: done, thanks

tyilo:
I just tested it on a project and it seems to work for me :)

Canop:
Do you see anything missing ?

And btw, do you have an idea how to get a reproducible order of items ?

PR #272, opened 2024-12-03
cpp analyzer

bryceberger:
There was a comment on the [reddit post] asking about C++ support, and I was somewhat interested in the internals of bacon.

This is a _very preliminary_ implementation of an analyzer that "parses" GCC and Clang warnings and errors. It pretty much just looks for the strings `"error: "` and `"warning: "`.

If anyone has interest in this, feel free to pick it up from here. I doubt I will bring this to a polished state.

There is an example cmake-based project included in the branch, with some rules added to the base `bacon.toml`.

image

[reddit post]: https://www.reddit.com/r/rust/comments/1h3d3ag/comment/lzpvbsv/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button


bryceberger:
> when testing changes on bacon, you don't have to modify bacon's own bacon.toml: you just need a bacon.toml in the test project directory, and you can for example do, when in bacon's dir, cargo run -- path/to/test-project

I tried that, but it didn't work. The docs on `Settings::read()` imply that it should, but reading the function I don't see where it checks the path. The following diff fixes it for me:

```diff
diff --git a/src/conf/settings.rs b/src/conf/settings.rs
index b137ff9..dd67fd1 100644
--- a/src/conf/settings.rs
+++ b/src/conf/settings.rs
@@ -108,6 +108,9 @@ impl Settings {
             Some(context.package_config_path()),
             Some(context.package_dot_config_path()),
             config_path_from_env("BACON_CONFIG"),
+            args.path
+                .as_ref()
+                .map(|p| PathBuf::from(p).join("bacon.toml")),
         ];
         for path in paths.into_iter().flatten() {
             if path.exists() {
```

bryceberger:
That doesn't work either. Pretty sure it's because there's a `Cargo.toml`, so `cargo metadata` finds the root. Then, `Context::package_directory` is set to `metadata.workspace_root` (instead of `intended_dir`, as it would if there wasn't a `Cargo.toml`), which is then used in `context.package_config_path()`.

Canop:
Hum. Yes, I hadn't considered the case of non-cargo projects when devising the path thing... The idea was that a directory without Cargo.toml inside another one couldn't be the package directory... We'll have to revise the logic.

Canop:
Regarding the analyzer, what should probably be done is what's done for ruff ( https://github.com/Canop/bacon/pull/273 ) so that we get a location to jump for.

bryceberger:
> We'll have to revise the logic.

c7e91fead56a05e17c4e14970ebfccd4f72552e0 uses the solution from https://github.com/Canop/bacon/pull/272#issuecomment-2515217769, will remove if/when there's another solution.

> Regarding the analyzer, what should probably be done is what's done for ruff (#273)

Assume this is talking about using `burp::...`. Done in e8e50e6731cd3445371f7eebb7efcc59322be038. Copied the implementation of `burp::error_line_ts()` from the referenced PR, added `burp::warning_line_ts()`. Should probably be rebased before merging.

Canop:
I hope to find the time this week-end to test, cleanly merge, etc.

Canop:
Working with a non cargo project inside a cargo project must be dealt separately from the "cpp analyzer" project, I made another issue: https://github.com/Canop/bacon/issues/274

Canop:
Merged.
I couldn't test the gcc_doctest  analyzer, though, can you give me an example config (not involving ninja).

Issue #224, opened 2024-10-04
[Feature] String search


crumblingstatue:
Sometimes I have a huge output, and I would like to be able to quickly find a lint/error/etc. of interest.

Example use case
- I enable `clippy::pedantic`
- There is tons of warnings, so I toggle summary view to find interesting lints
- I see an interesting lint that I would like to enable, but the summary view doesn't show the lint name
- I untoggle summary view, but that resets the scroll, so I'm not at the same lint anymore
- I need a way to find that lint again in the sea of output (by searching for part of the lint message), but there is no search feature

I reflexively reach for `/`, since it's a standard binding in a pager for string search, so that would be a nice default binding for it.

Canop:
> I reflexively reach for /, since it's a standard binding in a pager for string search, so that would be a nice default binding for it.

You can be confident I couldn't choose any other one ;)

Jirubizu:
is there any hope for a feature like this? I would 100% use it often :D

Canop:
Yes, there's "hope". There's just not enough free time in my days at the moment.

But upvoting this issue is helpful: it makes the issue goes up in the todo list as it make me realize that several people want it.

Canop:
@crumblingstatue @Jirubizu Can you please test https://github.com/Canop/bacon/pull/278 ?

Canop:
@crumblingstatue @Jirubizu Can you please check the PR #278 ?

Canop:
done

PR #278, opened 2024-12-16
search input

Canop:
Focus the search input with <kbd>/</kbd>.

When search input is focused, unfocus it with either :
* <kbd>esc</kbd>,  cancelling it
* <kbd>enter</kbd>,  freezing it

Navigate between matches with `tab` and `alt-tab`.

You can define other shortcuts, eg

```
[keybindings]
n = "next-match"
shift-n = "previous-match"
```

But those ones aren't ideal: they conflict with the <kbd>n</kbd> being currently bound to nextest, and they can't be used when the search is focused (you can hit <kbd>enter</kbd> before navigating but then you'll have to hit <kbd>/</kbd> again if you want to add to the pattern).

#224

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "bacon"
-version = "3.6.0"
+version = "3.6.0-dev"
 dependencies = [
  "anyhow",
  "cargo_metadata",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "bacon"
-version = "3.6.0-dev"
+version = "3.6.0-dev-search"
 dependencies = [
  "anyhow",
  "cargo_metadata",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bacon"
-version = "3.6.0"
+version = "3.6.0-dev"
 authors = ["dystroy <denys.seguret@gmail.com>"]
 repository = "https://github.com/Canop/bacon"
 description = "background rust compiler"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bacon"
-version = "3.6.0-dev"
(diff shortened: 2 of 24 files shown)


Canop:
I'm still not sure of what would be best: filtering items or lines. Current version is filtering lines but it's probable items would be better.

crumblingstatue:
> I don't think a filtering logic is the right one, there, I'll rather make it find and focus the right parts in the output

I agree

> enable navigating between matches with tab and alt-tab.

I don't think alt-tab is a good idea, as it's a standard key combo for switching windows, making the terminal window unfocus.

The `less` pager uses `n` and `shift+n` for next/previous.

Canop:
I think I have the right behavior now.

I still need to cover the case of a search match spanning over the line wrapping location, but the rest seems OK.

Jirubizu:
Had a quick view at MR (currently only have my phone available) unless I missed it. This is just doing basic string searching right? So for example regex patterns would not work?

Btw thank you for getting around to doing this feature!

Canop:
> This is just doing basic string searching right? So for example regex patterns would not work?

Correct.

crumblingstatue:
I think it would be nice to update the contextual help when in search mode.
The standard `Hit s to toggle summary mode, ? for help, w to not wrap lines, p to pause, q to quit` does not apply, since all the keys are captured by the search bar.
It could instead show `Hit Tab for next result, Shift+Tab for previous`.

Canop:
I wondered about that, then thought everybody would think about those. But I'll add them.

Canop:
Done
image

Canop:
I missed something in your remark. It's exact that some keyboard shortcut shouldn't be mentioned when the input is focused because they wouldn't work...

crumblingstatue:
The search offsets are wrong in summary mode. Hopefully it's not too hard to make it work properly for both normal and summary mode.

crumblingstatue:
One possible cool improvement would be to show the number of results, and also say "no results" if there are no results.
The bottom row is pretty busy, but maybe it could be added to the top indicator row?
image

Canop:
image

crumblingstatue:
Thank you, it seems to be working well!
Although maybe it would make it more clear to display "No results" or similar when there are no results for the search input, rather than display nothing.

crumblingstatue:
Thank you, the feature seems to be working well, but I'll do some more testing today. It seems to be ready for the most part though.

Canop:
I'll merge it. Related fixes and improvements, if necessary, will come in later PR.

c-git:
This feature is really cool.

PR #286, opened 2024-12-25
Show an error if the command fails to spawn

jyn514:
Before, bacon showed no output at all if spawning failed. Now, it gives a useful error instead:
```
Error: error in computation for job 'run'

Caused by:
    0: failed to spawn cd "../example" && CARGO_TERM_COLOR="always" RUST_BACKTRACE="0" "./run.sh"
    1: Exec format error (os error 8)
```

To test this, you can create a bacon.toml with `jobs.run.command = ["/not/here"]` and execute `bacon run`.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -153,9 +153,8 @@ fn run_mission(
                             action = state.action();
                         }
                         CommandExecInfo::Error(e) => {
-                            warn!("error in computation: {}", e);
                             state.computation_stops();
-                            break;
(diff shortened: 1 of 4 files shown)


Canop:
This is indeed much better.
Are you going to add anything to the PR or is it ready?

jyn514:
it's ready, if i make more changes i'll add them in followups.

i noticed that the "run" output is often hidden if the command succeeds, because it doesn't trigger the case in my second commit and it doesn't look like cargo output. but if i change that i'll fix that in a followup, not here.

Issue #282, opened 2024-12-22
Add an action copying the whole raw output


Canop:
This action putting the content of the "raw" output in the clipboard, when bound to a key, would solve the need mentioned in https://github.com/Canop/bacon/issues/240#issuecomment-2558216741

PR #290, opened 2025-01-05
Add ability to copy output to clipboard via keybinding

letsgetrusty:
Solves #282 

Note: Not sure if `ctrl + o` is the right keybinding. Open to suggestions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,7 @@ dependencies = [
  "clap",
  "clap-help",
  "cli-log",
+ "copypasta",
  "crokey",
  "ctrlc",
  "directories-next",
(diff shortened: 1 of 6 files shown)


Canop:
I'll be cautious here before adding such a dependent-heavy crate in bacon, especially one with native dependencies.

There are also alternatives with different trade-offs than copypasta (which eg doesn't support termux).

letsgetrusty:
I’ve switched from `copypasta` to `arboard`. Reasoning:

1. `arboard` has about 2x fewer dependencies than `copypasta` and is maintained by the 1Password team, ensuring long-term support and reliability.
2. I reviewed alternatives on [crates.io’s clipboard page](https://crates.io/keywords/clipboard) and didn’t find any better options.
3. While `arboard` doesn’t support `termux`, I believe this use case is niche enough that it’s not worth worrying about.
4. Rolling a custom clipboard manager introduces complexity and make `Bacon` harder to maintain.

I think clipboard functionality is common enough to include by default. That said, we could gate this functionality behind a feature flag.

Let me know your thoughts!

Canop:
The reason I mention termux is that it's a requirement I often have (I made [terminal-clipboard](https://github.com/Canop/terminal-clipboard) for that), but I don't think bacon has much chance of being used on termux short and medium term.

Any other opinion on the right choice of a clipboard crate is welcome.

Required:
* X11 support
* Mac OS support
* Windows support
* clean enough lib

Nice to have:
* Wayland support (without going through the X11 protocol)
* no loss of the clipboard on app exit
* not too many dependencies

**Not** required:
* support for bytes, images, etc.

I'll take some time before merging this so that this choice can be discussed (and so that I can test). 

A detail: maybe call it `copy-unstyled-output` to be clearer and let some room for future additions ?

PR #294, opened 2025-01-09
headless mode

### LINKED ISSUE #293: Support for non-interactive terminals and containers
Many of us develop in non-interactive terminals. Particularly those of us who work with embedded programming, or run dev stacks in containers, do not have the luxury of a text-based interactive UI. However, we'd still love a job runner that responds to filesystem changes.

It's important that bacon can be run in a mode where its only output is to stdout/stderr, it expects no input, and all it does is just runs the default job. Just a boring, colorless task runner to be run from a container.

Canop:
When launched with `--headless`, bacon doesn't render the TUI and doesn't accept user events.

In this mode, it just displays the command output followed by a one line analysis:

image

Command output lines coming on stdout are rendered on stdout. If from stderr, they go to stderr.

Fix #293

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,7 @@ dependencies = [
  "clap-help",
  "cli-log",
  "crokey",
+ "ctrlc",
  "directories-next",
  "gix",
  "glob",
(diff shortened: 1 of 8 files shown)


Canop:
It would also be quite trivial to remove, on option, style related escape sequences.

sssilver:
I ran this in headless mode in a container and ran into a few questions. Might be better to talk in real time? Do you use Discord or IRC or literally any real-time messaging platform?

Canop:
I'm on Miaou in the broot room: https://miaou.dystroy.org/3490

crisidev:
This is a great feature. I would like to allow bacon to be started by bacon-ls to ease up that usecase!

Canop:
From @sssilver trial, it looks like bacon, when running in docker on mac, doesn't detect file events occuring in a volume backed by the host. IMO this shouldn't prevent merging this but I take opinions.

crisidev:
It seems to be somehow a widespread issue with OSXFS: https://docker-sync.readthedocs.io/en/latest/troubleshooting/sync-stopping.html

Canop:
I merged.

But feedback is still welcome, and I'll change how it works if necessary.

PR #295, opened 2025-01-10
Allow bacon to be consumed as a nix flake

crisidev:
I'd like to be able to consume bacon as a nix flake if you agree on adding the support.

If you don't use nix and don't want this to be part of bacon, I am happy to continue maintaining the flake or add it to bacon-ls flake.

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bacon.log
 /website/site
 .bacon-locations
 bacon-analysis.json
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,94 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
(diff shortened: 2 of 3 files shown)


crisidev:
This would probably require documentation, but the required code is more or less this

crisidev:
> > What do you think?
> 
> This looks good. What documentation do you think should be added.
> 
> I'd welcome comments from anybody else with nix sapience.

I guess how to use this in home-manager and nixos? Not sure thou..

crisidev:
The flake also provides an overlay, so using it in home-manager is pretty easy, you just have to follow the home-manager overlay documentation. For nixos, I have no clue because I don't use it.

crisidev:
Feel free to merge it if this is enough for you

alerque:
My take is that this doesn't need any documentation. Anybody that knows how to use a Flake at all will know how to use this, and no amount of documentation from the project side is going to be "enough". For me just knowing this PR exists and looking at the source branch I was able to run the Flake as `nix run github:crisidev/bacon/main` (later when this PR merges it will be `nix run github:Canop/bacon`, but if folks don't have `nix` setup to run Flakes at all that can be a bit of a process, and documenting it shouldn't be up to the project.

PR #296, opened 2025-01-11
--config-toml argument

### LINKED ISSUE #284: Pass the command for a job as an argument to bacon
This is by no means a need but I've look and can't see that anyone asked before. I wanted to know if supplying the command job details to run on the command line would be something you'd be amenable to. So you could run one off jobs for example (I know cargo test has a built in job just didn't have a better idea for the example).

```sh
bacon -c "cargo test"
```

would be equivalent to having a `bacon.toml` like this:

```toml
[jobs.one-off]
command = ["cargo", "test"]
```

and then running

```sh
bacon -j one-off
```

Canop:
Fix #284

Feedback welcome

### DIFF
--- a/src/conf/args.rs
+++ b/src/conf/args.rs
@@ -126,6 +126,10 @@ pub struct Args {
     #[clap(long, value_name = "project")]
     pub project: Option<String>,
 
+    /// Configuration passed as a TOML string
+    #[clap(long)]
+    pub config_toml: Option<String>,
+
     #[clap()]
     /// What to do: either a job, or a path, or both
     pub args: Vec<String>,
--- a/src/conf/settings.rs
+++ b/src/conf/settings.rs
@@ -88,6 +88,7 @@ impl Settings {
     /// * the package level `bacon.toml` file in package-root/.bacon.toml
     /// * the package level `bacon.toml` file in package-root/.config/.bacon.toml
     /// * the file whose path is in environment variable `BACON_CONFIG`
+    /// * the content of the `--config-toml` argument
     /// * args given as arguments, coming from the cli call
     pub fn read(
         args: &Args,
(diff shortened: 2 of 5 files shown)


c-git:
Going to test now, I really love how nicely you integrated the idea into the code. It feels obvious what to expect and I think that's desirable. Regarding your [comment on the downside](https://github.com/Canop/bacon/issues/284#issuecomment-2585312117)

> A downside of the chosen solution is that if you specify a job, and it's not the default job, it's not the one executed, which may be surprising for who isn't used to the multi job system of bacon.

I guess only time will tell if that actually materializes as a problem but my thinking is that this is not likely to be a problem. Most of the time people will be running one liner commands they found in tutorials or documentation not writing it by hand. And it's pretty easy to identify the problem if they provide the command.

c-git:
Once we finalize we'll also need to include an example on the website. Just mentioning it before I forget.

It works as expected but there are a few gotchas that I found while trying it out. 
- You need to wrap the TOML in single quotes otherwise you need to escape all the double quotes.
- TOML needs white space so you must use a multi line command.

Those are just statements to capture what I observed but I still think this provides the most flexibility and adds the ability to share commands that people can copy and paste. It allows you to iterate on your job in the bacon.toml then copy and paste it into a command line command. Overall I'm happy with it.

This is my test command that did indeed work as desired

```sh
bacon -j cli-test --config-toml '
[jobs.cli-test]
command = [
  "sh",
  "-c",
  "echo \"hello $(date +%H-%M-%S)\"; cargo run",
]
need_stdout = true
allow_warnings = true
background = false
on_change_strategy = "kill_then_restart"
#kill = ["pkill", "-TERM", "-P"]'
```

Canop:
I think I'll just add your example in the cookbook page.

c-git:
That's fair but it' is something you can copy and paste and just try without multiple steps.

Canop:
> That's fair but it' is something you can copy and paste and just try without multiple steps.

Yes. I was just wondering how to call that thing...

c-git:
Yeah maybe "bacon inline command snippet" ? But that does seem long.

PR #297, opened 2025-01-12
add alias to job configuration

tversteeg:
This allows you to write:

```toml
[jobs.example]
alias = ["ex"]
command = []
```

So you can do both `bacon example` and `bacon ex`.

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -38,6 +38,7 @@ command = [
 need_stdout = false
 
 [jobs.win]
+alias = ["windows"]
 command = [
 	"cross", "build",
 	"--target", "x86_64-pc-windows-gnu",
--- a/src/conf/settings.rs
+++ b/src/conf/settings.rs
@@ -249,10 +249,29 @@ impl Settings {
             bail!("Invalid configuration : no job found");
         }
         if let NameOrAlias::Name(name) = &self.default_job.name_or_alias {
-            if !self.jobs.contains_key(name) {
+            if self.job(name).is_none() {
                 bail!("Invalid configuration : default job ({name:?}) not found in jobs");
             }
         }
         Ok(())
+    }
+
+    /// Get a job by its name or alias.
+    pub fn job(
+        &self,
+        name: &str,
+    ) -> Option<&Job> {
+        self.jobs.iter().find_map(|(key, job)| {
+            if name == key
+                || job
+                    .alias
+                    .as_ref()
+                    .is_some_and(|alias| alias.iter().any(|alias| alias == name))
+            {
+                Some(job)
+            } else {
+                None
+            }
+        })
     }
 }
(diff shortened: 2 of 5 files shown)


Canop:
Is that really needed ?

This raises a question for users: when defining a job so as to replace another one, which is frequent, do you also automatically replace the aliases? Or do you have to specify the same set of aliases ?

tversteeg:
If it's really needed is of course highly subjective, but for my use cases it would be a nice feature.

> This raises a question for users: when defining a job so as to replace another one, which is frequent, do you also automatically replace the aliases? Or do you have to specify the same set of aliases ?

I would say that it replaces the aliases, something like this:

```toml
[jobs.overridable]
alias = ["old", "over"]
command = ["new"]
```

```toml
[jobs.overridable]
alias = ["new", "over"]
command = ["new"]
```

So you get the following results:

| CLI | Result |
| -- | -- |
| `bacon overridable` | `new` |
| `bacon over` | `new` |
| `bacon new` | `new` |
| `bacon old` | `old` |

Canop:
I wasn't clear.

Imagine there's a team specific job "check" with alias "c", and a key-binding `c = "job:check"`. Some team member get used to `bacon c`.

Now, on a project of this team, a `bacon.toml` defines a new `"check"`job because it uses special features maybe. Most members of the team will probably be surprised that `bacon check` and `bacon c` no longer do the same thing.

tversteeg:
The `c` shortcut not calling the new `"check"` job is not related to this PR, right? 

If the new `"check"` job doesn't add the `"c"` alias, `bacon check` would work as intended and `bacon c` would throw an error that the job could not be found. I think this behavior makes sense.

Another option would be to merge all aliases from each previous configuration in the new job.

Yet another option would be to keep the previous job definitions and still allow their aliases to be called if they are not overwritten in the new definition.

Canop:
Other user opinions welcome

c-git:
Having given this a bit more thought, I think with the use of global jobs. The extra typing might be a one time effort and adding diminishes the value of adding aliases in my opinion. @tversteeg do you normally have different aliases per project or are they just things you could add to prefs.toml?

tversteeg:
> If I did need aliases I think I would create a job for each alias instead of defining them in the job I want aliased (Example below). I think that would make it more obvious what priority they have based on where they are defined. It is more typing so if it were something I used often I could see how I would want the syntactic sugar instead.
> 
> ```toml
> [jobs.example]
> command = ["cargo", "run", "--example"]
> need_stdout = true
> allow_warnings = true
> 
> [jobs.ex]
> command = ["echo", "alias used"]
> on_success = "job:example"
> 
> [jobs.exs]
> command = ["echo", "alias used"]
> on_success = "job:example"
> ```

That's actually a good workaround that would also fulfill my current needs.

> Do you normally have different aliases per project or are they just things you could add to prefs.toml?

I have different aliases per project, not globally.

PR #302, opened 2025-01-15
fix middle click sending bacon to outer heaven 

### LINKED ISSUE #298: middle click sends bacon to outer heaven
```
bacon 3.7.0
in tmux 3.5a
in kitty 0.38.1
```

action taken: click the middle mouse button

expected result: nothing happens

actual result: bacon cycles through the available job types several times like some sort of wheel of fortune, before stopping on the `test` job, which it then executes.

is this intended behavior?

Canop:
Fix #298 

The problem of middle-click is that it's used on linux (especially linux terminals) to send as key events the content of the clipboard.

This PR "solves" the problem by making bacon manage mouse events by itself.

With this PR, bacon scrolls up and down using the mouse wheel events.

Downsides:
- potential regression for users pasting with middle click into the search input
- **major** regression for users selecting and copying with the mouse
- not using the scroll multiplier defined OS wide

Given the downsides, this PR can't be merged right now.
I would need at least to manage selecting (which could be better that today's selection as it would handle the scrollbar correctly), but also replace the menu for copying that most terminals propose on right click.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### next
 - with `--headless`, bacon runs without TUI - Fix #293
 - `--config-toml` argument - Fix #284
+- fix workspace level Cargo.toml file not watched
 
 <a name="v3.7.0"></a>
 ### v3.7.0 - 2024/12/27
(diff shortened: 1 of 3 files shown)


Canop:
Managing a menu on right click, selecting with the mouse, this looks quite feasible. But some people probably use other mouse related features, eg other menu enabled features that would be specific to their terminal application.

PR #307, opened 2025-01-17
Avoid freezes when reloading configs on macos

### LINKED ISSUE #306: Bacon freezes when editing bacon.toml
Editing `bacon.toml` while running bacon (latest release or from `main`) causes bacon to freeze, and the only remedy I've found is to kill the process.

I'm on macOS 14.7.1.

From `bacon.log`:

```
12:28:55.121 [INFO] bacon::app: config watch event received
12:28:55.121 [DEBUG] bacon::app: requested action: Internal(ReloadConfig)
12:28:55.121 [INFO] bacon::watcher: notify event: Event { kind: Create(File), paths: ["/Users/ian/dev/koto/koto/bacon.toml"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }
12:28:55.121 [INFO] bacon::watcher: ignorer.excludes_all_pathbufs(&we.paths) took 291ns
12:28:55.121 [DEBUG] bacon::watcher: at least one is included
12:28:55.121 [DEBUG] bacon::exec::executor: failed to send 'die' signal: sending on a disconnected channel
```

irh:
On macos reloading the config file results in a freeze, caused by <https://github.com/notify-rs/notify/issues/208>.

The freeze occurs when the watchers in `run_mission` are dropped while exiting from the function.

Reordering the fields so that `_notify_watcher` is dropped after `receiver` fixes the issue, although I'm unsure why.

At first I thought that the channel was being used by the notify watcher, which is why I tried the field reordering, I thought there was a dependency creating a deadlock. This isn't correct though; the notify watcher has its own thread that is being used for processing events, so I don't understand the interaction.

That said, it appears to be a reliable fix. Without this change the freeze occurs consistently, but with this change reloading the config never freezes.

Fixes #306.

### DIFF
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -22,8 +22,8 @@ use {
 
 /// A file watcher, providing a channel to receive notifications
 pub struct Watcher {
-    _notify_watcher: RecommendedWatcher,
     pub receiver: Receiver<()>,
+    _notify_watcher: RecommendedWatcher,
 }
 
 impl Watcher {
@@ -88,8 +88,8 @@ impl Watcher {
             }
         }
         Ok(Self {
-            _notify_watcher: notify_watcher,
             receiver,
+            _notify_watcher: notify_watcher,
         })
     }
 }


Canop:
I'll probably release 3.8.1 tomorrow with this fix.

PR #311, opened 2025-01-19
Wait grace_period before reloading config

### LINKED ISSUE #310: Race condition in config reloading
When reloading the config file, it is possible that `bacon` exits with `Error: job not found: "<current_job>"`.

Reproduction:
```console
$ cat bacon.toml
default_job = "echo"

[jobs.echo]
command = ["echo", "1"]
need_stdout = true
$ bacon echo
[...]
```
When saving the config file in e. g. vim, neovim or helix (but not when using `touch` or saving in gedit) there is a chance that bacon will exit with
```
Error: job not found: "echo"
```
This usually happens within a few tries, but usually takes longer when `BACON_LOG=debug` is specified.

I found this while trying to reproduce a hang on config reload similar to #306 with a small config file. With a normal sized config file, I have never seen this.

Given that this is editor-dependent, behaves differently on a `tmpfs` filesystem, and happens somewhat randomly and more likely on smaller config files, I think this issue is caused by the same editor behaviour as #222.

bacon version: a66fa35d8d619d3c7bf38455301a7d3b244ef788
OS: Linux
No `~/.config/bacon/prefs.toml` file.

Canop:
Fix #310

### DIFF
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -27,7 +27,7 @@
 # moves the file to a backup name before recreating the right one)
 # You can set it to "none" if it's useless for you.
 #
-grace_period = "2ms"
+# grace_period = "5ms"
 
 
 # Uncomment and change the value (true/false) to
--- a/defaults/default-prefs.toml
+++ b/defaults/default-prefs.toml
@@ -22,11 +22,12 @@
 
 # The grace period is a delay after a file event before the real
 # task is launched and during which other events will be ignored.
-# This is useful if several events are often sent quasi-simultaneously
-# (eg your editor backups before saving, then renames).
+# This is mostly useful when your editor does several operations
+# when saving a file and the state is temporarly wrong (eg it
+# moves the file to a backup name before recreating the right one)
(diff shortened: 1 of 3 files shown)


narpfel:
This works for me.

This also fixes (or works around?) the deadlock I was initially trying to reproduce when I stumbled upon #310: It turns out that `git` also does multiple delete and write operations in quick succession when e. g. switching branches or when stashing. For me, this often caused bacon to completely deadlock.

Thanks!

Canop:
I never encountered this one either. Maybe the filesystem type is involved. What's yours ?

In any case I decided to set the default value of the grace_period to 2ms rather than none: there's no real problem in waiting a little.

narpfel:
> Maybe the filesystem type is involved. What's yours ?

I can reproduce #310 on btrfs (SATA SSD and M.2 SSD) and on ext4 (HDD) across two different computers and on my Raspberry Pi (btrfs on an SD card).

On a tmpfs I can only reproduce it on the Pi and the slower of the two computers.

Thanks again for the quick fix!

PR #314, opened 2025-01-22
Make sure we can build the sound support with Nix

crisidev:
Bacon now requires pkg-config and alsa-lib to compile with Nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,6 +21,7 @@
 
         naersk' = pkgs.callPackage naersk { };
         bacon = naersk'.buildPackage {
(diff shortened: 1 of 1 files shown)


Canop:
It's needed when compiled with the "sound" feature, right ?

Is this feature a problem for nix ?

crisidev:
By default this nix flake compiles with all features, so yeah, it is required right now. I can work on a fix next week to select features.

Canop:
I don't know if it's needed. I don't have feedback relative to the features choice (eg compilation problems). So right now, we may wait before engaging into that.

PR #316, opened 2025-01-23
[Nix] Make sure we pull alsa and pkg only on Linux

crisidev:
The dependency `rodio` requires alsa and pkgconfig only on Linux. My previous commit was breaking Nix build on other platforms 😞 

Apologies for the oversight.


Canop:
> Apologies for the oversight.

No worry. This kind of things happen.

PR #320, opened 2025-01-28
All jobs config

Canop:
This refactors job settings:

* allow some settings which were only allowed at root to be also at the job level (for example grace_period, or sound)
* allow some settings which were only allowed inside jobs to be also at the root
* remove the duplication of settings which were defined both in `Job` and in `Config` (and sometimes also in `Settings`)

Now, *all* parameters of a job can also be defined at root level (which means that a job definition can theoretically be empty).

### DIFF
--- a/src/jobs/job.rs
+++ b/src/jobs/job.rs
@@ -33,8 +33,7 @@ pub struct Job {
 
     /// The tokens making the command to execute (first one
     /// is the executable).
-    /// This vector is guaranteed not empty
-    /// by the PackageConfig::from_path loader
+    #[serde(default)]
     pub command: Vec<String>,
 
     /// Whether to apply the default watch list, which is
@@ -46,6 +45,7 @@ pub struct Job {
     pub default_watch: Option<bool>,
 
     /// Env vars to set for this job execution
+    #[serde(default)]
     pub env: HashMap<String, String>,
 
     /// Whether to expand environment variables in the command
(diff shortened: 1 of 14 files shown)


c-git:
Wow this would be great. Let me know when you're ready to have people test it. I'd be happy to try it out.

Canop:
I think you can test it :)

Canop:
That would be great. Honestly I'm not yet confident I didn't mess a fundamental thing here and I made almost no test myself.

c-git:
In that case I'll try to look at the code first instead of testing. It would be easier for me to find any possibly unintentional things that way than with testing which only shows me one path.

src/jobs/job.rs
```diff
@@ -129,23 +124,104 @@ impl Job {
...
+        }
+        if let Some(b) = job.need_stdout {
+            self.need_stdout = Some(b);
+        }
+        if let Some(v) = job.on_change_strategy {
+            self.on_change_strategy = Some(v);
+        }
+        if let Some(v) = job.on_success.as_ref() {
+            self.on_success = Some(v.clone());
+        }
```
c-git, line 215: There doesn't seem to be code to merge grace_period

c-git (commented):
Didn't get a chance to test like I'd hoped but was able to review the code at least for now. It looks good. Really nice cleanup. I like how you took the oppotunity to simplify the code.

src/jobs/job.rs
Canop: Good find!

I'll add a unit test creating 2 instances of Job, without using the Default implementation, to check there's nothing missing. Or I'm sure I'll add regressions later on...

Canop: In another project I use macros for the same exact cases, but that other project takes minutes to compile ^^

Canop: I fixed it, with the unit test. There might still be traps, eg on not being able to set some default-looking values.

c-git:
All my tests seem to have worked as expected. Got a bit tripped up with errors due to sound config but eventually figured it out and created an issue. Thank you very much for this feature. Going to clean up my config files now to remove the duplication I added for the sound.

c-git:
I don't recall if we mentioned the no-op in the documentation but that would also need to be updated

PR #325, opened 2025-02-11
implement dynamic completions for jobs

bryceberger:
Implements dynamic completions using <https://docs.rs/clap_complete/latest/clap_complete/>. Use with, e.g. `COMPLETE=fish bacon | source`.

Allows you to complete job names. Example: `bacon <TAB>` ->
image

Currently marked as draft because this does not resolve directory arguments, i.e. `bacon subdir/ <TAB>` will *not* read `subdir/bacon.toml`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "clap-help",
+ "clap_complete",
  "cli-log",
  "crokey",
  "ctrlc",
@@ -389,6 +390,18 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+]
+
+[[package]]
+name = "clap_complete"
+version = "4.5.44"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "375f9d8255adeeedd51053574fd8d4ba875ea5fa558e86617b07f09f1680c8b6"
+dependencies = [
+ "clap",
+ "clap_lex",
+ "is_executable",
+ "shlex",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Canop:
Can you please give me pointers on this `COMPLETE=fish bacon | source` thing, how it is used ?

bryceberger:
`COMPLETE=fish bacon` will output a fish script that installs dynamic completions for fish. Sourcing that will then install the completions for the current shell session. An example for how this works:


```bash
# you type:
$ bacon c<TAB>

# fish invokes:
$ COMPLETE=fish bacon COMPLETE=fish bacon -- c c
# which outputs...
check-all
check
clippy-all
clippy
```

(A similar process happens for bash and zsh)

Essentially, instead of a static list of completions, the shell will invoke the binary behind the users back and ask _it_ to provide relevant completions. `clap_complete` deals with the format, so bacon just has to provide `ArgValueCandidates` for relevant fields.

Canop:
ok. This has to be adapted a little depending on the shell. On my system I must send the script to a file because `source` needs the file as argument. But it works.

Now, there's a problem that won't be obvious for users: this command should be ran every time you go to a project (and ideally every time the bacon.toml file changes) in order to have up to date jobs.

bryceberger:
Yeah, bash (and zsh?) need to `source <(COMPLETE=bash bacon)`, because `source` on those shells has to have a file.

The completions only need to be sourced once, on shell startup. After that, it will generate the correct completions in any directory, regardless of updates to `bacon.toml`.

You can easily see what completions will be generated in any directory by doing `bacon --completion-list-jobs` to print a null-separated list.

Canop:
> The completions only need to be sourced once, on shell startup. After that, it will generate the correct completions in any directory, regardless of updates to bacon.toml.

Oh, that's right. That's great!

Canop:
Well, broot does add to the shell startup script on first run.

c-git:
Ah so you already have experience with this. Ok cool.

Canop:
@bryceberger What's your plan here ? Are you going to solve the problem you mentioned with the subdir thing, to you prefer me to look at that ?

bryceberger:
Apologies, forgot to respond to this. The above commit works well enough for me --- complete jobs when empty, complete project when specifically requested. Example:

```bash
bacon <TAB>
# list of jobs
# list of flags

bacon --job <TAB>
# list of jobs

bacon --project <TAB>
# list of directories

bacon --<TAB>
# list of flags
```

If you want to work on it from here, feel free.

Canop:
I merged.
Installing the sourcing line will be done later, depending on feedback on this completion.

PR #327, opened 2025-02-11
Detect `cargo test` stack overflows

### LINKED ISSUE #326: Failing test reported as passing given stack overflow, at least one passing test, and a compiler warning
Run `bacon test` on a cargo project containing the following code

```rust
fn a_warning() {}

#[test]
fn another_test() {}

#[test]
fn test() { 
    if 1 == 1 {
        test()
    }
}
```

It outputs

```
proj-name   test   pass!    1 warning   

1  warning: function `a_warning` is never used
--> src/lib.rs:1:4
 |
1 | fn a_warning() {}
 |    ^^^^^^^^^
 |
 = note: `#[warn(dead_code)]` on by default
```

Despite the fact that `cargo test` reports

```
warning: function `a_warning` is never used
 --> src/lib.rs:1:4
  |
1 | fn a_warning() {}
  |    ^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: `overflow` (lib) generated 1 warning
warning: `overflow` (lib test) generated 1 warning (1 duplicate)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests src/lib.rs (target/debug/deps/overflow-b0e0b11f646c1cda)

running 2 tests
test another_test ... ok

thread 'test' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `/tmp/overflow/target/debug/deps/overflow-b0e0b11f646c1cda` (signal: 6, SIGABRT: process abort signal)
```

Changing `1 == 1` to `1 == 2` makes the test pass, but doesn't change bacon's output at all since it reports it as it if was already passing.

gmorenz:
Fixes #326 by matching the stack overflow message.

Doesn't attempt to detect the `error` message to know that *something* went wrong in general, to prevent future similar misses. I think that would be a good idea, but I'll leave it to someone else.

### DIFF
--- a/src/analysis/standard/standard_line_analyser.rs
+++ b/src/analysis/standard/standard_line_analyser.rs
@@ -33,6 +33,9 @@ fn analyze_line(cmd_line: &CommandOutputLine) -> LineAnalysis {
         } else if let Some(k) = as_fail_result_title(content) {
             key = Some(k.to_string());
             LineType::Title(Kind::TestFail)
+        } else if let Some(k) = as_stack_overflow_message(content) {
+            key = Some(k.to_string());
+            LineType::Title(Kind::TestFail)
         } else if cmd_line.origin == CommandStream::StdErr
             && regex_is_match!("^error: ", content)
             && !regex_is_match!("^error: aborting due to", content)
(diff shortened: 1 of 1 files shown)


Canop:
This is a good improvement.

PR #336, opened 2025-03-06
Upgrade vte to 0.15

nc7s:
This is mainly for packaging bacon in Debian, where vte is 0.13. (0.15 only additionally requires the bytes change for `Parser::advance()`, so it's an easy patch.)

I'm not sure if the changes are right though, can't see from tests or running it, so a review is necessary and appreciated.

vte gained CSI subparameter support in [0.9](https://github.com/alacritty/vte/blob/master/CHANGELOG.md#090), passing [`vte::Params`](https://docs.rs/vte/0.15.0/vte/struct.Params.html) in place of old `[i16]`. `Params::iter()` gives an iterator (`ParamsIter`), each `Item` of which is a `[u16]`, consisting of the "main" parameter and its subparameters.

For `if *params == [0]`, I assumed it means there's only one parameter, of the value 0, so for the iterator form I changed it to check if there's only one item and it's `[0]`.

For `TString::push_csi()`, the old cold just plain writes all parameters in the order they were received, so I iterated through the new items and write their parameters in order as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,12 +151,6 @@ checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
-
-[[package]]
-name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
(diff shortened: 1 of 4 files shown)


nc7s:
A side question that's a bit off topic: is there plan to divide bacon code into multiple crates? If so, would they be structured as libraries that can be reused by other things? This information is needed to decide on its packaging style. Thanks.

Canop:
Err... No, there's no plan to cut bacon at the moment. Bacon is already build on many crates.

Why would you want to cut it ?

nc7s:
For example, splitting analyzers out to their own crates? It's a common scheme in the Rust ecosystem, like ripgrep, so I'm asking. Thanks for the answer.

Canop:
I don't forget this PR. As I don't want to review the changes in the vte crate, I'm currently checking your PR mostly by using the modified bacon in use in a few of my projects.

nc7s:
No worries ;) This is not meant to be a "reminder", but only to add the updated Cargo.lock I forgot. I can always patch it in Debian (actually I did), so there's no need to rush.

Also thanks for testing it out!

Canop:
merged

PR #340, opened 2025-03-08
feat!: Add "back-or-quit" and don't quit on "back"

### LINKED ISSUE #338: keybindings: an action like `back` that doesn't quit
As a vim user, I'm accustomed to hitting `Esc` lots of times to get back to the "default" state, and using some other shortcut to actually quit apps (`:q` doesn't really have an equivalent in bacon but `Ctrl-C` seems fine).

As far as I can tell, there's no way to configure this kind of behavior in Bacon today, since the `back` action will quit if triggered from the main job page. Would you be open to adding an action like `back-no-quit` which does the same thing as `back` except never quits the app? Then I could have a configuration like this to avoid my muscle memory from constantly quitting Bacon 😅 
```toml
[keybindings]
q = "quit"
esc = "back-no-quit"
```

I'd be willing to try submitting a PR if this seems like a reasonable request.

Thanks!

ian-h-chamberlain:
Closes #338 

This could be considered a "breaking" change since user-defined keybindings behave differently, as do the builtin defaults.

The new default "back" action no longer quits, and there is a new "back-or-quit" which restores the old behavior.

This also by nature makes the same change for job refs: "previous" no longer quits automatically but "previous-or-quit" does.

### DIFF
--- a/src/jobs/job_stack.rs
+++ b/src/jobs/job_stack.rs
@@ -1,9 +1,6 @@
 use {
     crate::*,
-    anyhow::{
-        Result,
-        anyhow,
-    },
+    anyhow::anyhow,
 };
 
 /// The stack of jobs that bacon ran, allowing to get back to the previous one,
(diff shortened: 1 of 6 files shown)


src/jobs/job_stack.rs
```diff
@@ -45,7 +48,16 @@ impl JobStack {
                         }
                     }
                     None => {
-                        return Ok(None);
+                        if let JobRef::PreviousOrQuit = job_ref {
+                            return Err(None);
+                        }
+
+                        // this has a side effect of reloading the job, which isn't ideal
+                        // but the main loop needs some mission to avoid spinning forever
```
ian-h-chamberlain, line 56: This is the main caveat for this implementation; every time I tried to make it avoid reloading here, the result was the main `loop` constantly trying to pick a job and never running any mission, which freezes the application (can't even quit with e.g. ctrl-c and need to kill from another session).

If you can think of a way to make it so this _doesn't_ reload the same job again that would be great!

PR #341, opened 2025-03-09
Implement custom sounds

### LINKED ISSUE #337: Custom sounds
Currently, we have some sounds hand picked by the authors baked into the compiled executable. It works, but has shortcomings. In Debian particularly, the included sound resources have too many copyright holders: even though they are OK license wise with CC0, we can't reliably attribute each to their respective author(s). I'll have to exclude them, and disable the sound feature for now. Allowing custom sounds won't solve the attribution problem, but does enable the feature.

Currently sounds are referred to by names, e.g. `on_success = "play-sound(name=90s-game-ui-6,volume=50)"`. Presumably we can also refer by paths, e.g. `play-sound(path=~/Audio/my-awesome-notification.mp3,volume=42)`.

We could also allow naming custom sounds, something like this:

```toml
# prefs.toml
[sound.collection]
my-awesome-notification = '~/Audio/my-awesome-notification.mp3'
```

```toml
# bacon.toml
[jobs.check]
on_success = 'play-sound(my-awesome-notification,volume=42)'
```

nc7s:
Resolves #337.

Each commit is self contained (hopefully) and comes with explanation in its message (mostly). But to sum it up, notable changes are

- A new `[sound.collection]` config section to specify custom sounds, which can be used as `play-sound(name=custom-name)`; section name could be adjusted
- A `default-sounds` feature to allow removing, well, default sounds
- `play-sound(path=...)` for "one-off" settings

### DIFF
--- a/src/sound/sound_config.rs
+++ b/src/sound/sound_config.rs
@@ -2,6 +2,7 @@ use {
     crate::*,
     schemars::JsonSchema,
     serde::Deserialize,
+    std::collections::HashMap,
 };
 
 /// Sound configuration.
@@ -12,6 +13,7 @@ pub struct SoundConfig {
 
     /// Base volume, acting as a multiplier for the volume of specific sounds.
     pub base_volume: Option<Volume>,
+    pub collection: Option<HashMap<String, String>>,
 }
 
 impl SoundConfig {
@@ -24,6 +26,26 @@ impl SoundConfig {
         }
         if let Some(bv) = sc.base_volume {
             self.base_volume = Some(bv);
+        }
+        #[cfg(feature = "sound")]
+        // Load sounds configured in `[sound.collection]`.
+        // This doesn't "apply" this item in the sense other items are applied;
+        // but rather, add them to `super::play_sound::SOUNDS` so they can be
+        // looked up later.
+        // Ideally, they should be loaded at some more suitable point, or even
+        // passed to `super::play_sound::play_sound()` to achieve some degree
+        // of on-demand loading. But current structure of `SoundPlayer` makes
+        // that difficult.
+        if let Some(ref collection) = sc.collection {
+            let mut self_collection = self.collection.get_or_insert_default();
+            for (name, path) in collection {
+                if !self_collection.contains_key(name) {
+                    if let Err(e) = crate::sound::play_sound::add_sound(name, path) {
+                        warn!("failed to add sound(name = {name}, path = {path}): {e:?}");
+                    }
+                    self_collection.insert(name.clone(), path.clone());
+                }
+            }
         }
     }
     pub fn is_enabled(&self) -> bool {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,6 +168,7 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
+ "shellexpand",
  "termimad",
  "toml",
  "unicode-width 0.2.2",
@@ -731,6 +732,27 @@ checksum = "339ee130d97a610ea5a5872d2bbb130fdf68884ff09d3028b81bec8a1ac23bbc"
 dependencies = [
  "cfg-if",
  "dirs-sys-next",
+]
+
+[[package]]
+name = "dirs"
+version = "5.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
+dependencies = [
+ "dirs-sys",
+]
+
+[[package]]
+name = "dirs-sys"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
+dependencies = [
+ "libc",
+ "option-ext",
+ "redox_users",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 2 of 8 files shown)


src/sound/sound_config.rs
```diff
@@ -20,6 +22,22 @@ impl SoundConfig {
...
+        // looked up later.
+        // Ideally, they should be loaded at some more suitable point, or even
+        // passed to `super::play_sound::play_sound()` to achieve some degree
+        // of on-demand loading. But current structure of `SoundPlayer` makes
+        // that difficult.
+        if let Some(ref collection) = sc.collection {
+            for (name, path) in collection {
+                // TODO: Do we really want to panic here? Or should it be
+                // silenced?
+                crate::sound::play_sound::add_sound(name, path).unwrap();
```
Canop, line 38: Why would we want to have a panic ? That's not how bacon should behave

nc7s: Indeed, hence the TODO.

Do we silence it, or warn about it? If warn, where should the warning go?

nc7s: Indeed, hence the TODO.

Do we silence it, or warn about it? If warn, where should the warning go?

nc7s:
Hmm, I seem to have mistakenly assumed it's OK to only listen for interrupt if the duration is known. The sound is then not played at all, not yet sure why.

But instead of trying to "fix" acquiring duration, I think it's reasonable to just set a maximum duration: it's a notification sound, after all, it shouldn't last longer than a few seconds.

c-git:
> Hmm, I seem to have mistakenly assumed it's OK to only listen for interrupt if the duration is known. The sound is then not played at all, not yet sure why.
> 
> But instead of trying to "fix" acquiring duration, I think it's reasonable to just set a maximum duration: it's a notification sound, after all, it shouldn't last longer than a few seconds.

I use the default sounds and have no issue but when I used to do custom sounds (in the previous framework I used when I was using python) they were several seconds long. It was actually an audio report of the outcome. It would tell me if it passed or failed and the error count. I know the sounds won't be doing that but just pointing out that if we allow custom duration is not necessarily safe to assume it will be short. I haven't had time to look at the code myself. Is it possible to send to a separate thread or something like that if it's blocking something else so the user can choose any sound they like? If it's too complicated maybe it's not worth it but I just wanted to point out that I don't think it's a safe assumption. Alternatively we could ask the use for a max duration to play from the sound. Maybe it's their favourite song and they don't want to crate a cropped version but they don't actually want the whole 5 minutes.

nc7s:
> Alternatively we could ask the use for a max duration to play from the sound.

Sure could do, add an option `duration` to `[sound.collection]` entries and maybe the `play-sound` action.

> Maybe it's their favourite song and they don't want to crate a cropped version but they don't actually want the whole 5 minutes.

Agreed, which is basically the source of my idea of a default duration/cutoff. But then again,

> not necessarily safe to assume it will be short

It's a _notification_ sound, when a job is done - there is a reason why the pre-picked sounds are short, with the longest being 2 seconds (`2`, `success`) and the shortest being 0.25 seconds (`store-scanner-beep`), and the longer ones mostly just have some hardly audible fade-in and -out. Notifications happen a lot. _A lot_. Sounds gonna overlap if long yet not cut off. There's gonna be cutoff anyway, so why not set a "standard" one? Ability to set custom duration could be offered if the user feel like it, as said above.

I could even go as far as suggesting to replace mp3 with ogg - mp3 is not an open format, even if its patents expired; compared, ogg (container) and vorbis/opus (audio stream) are open formats.

.. But this all comes from the lack of assured ability to measure duration. It's baffling symphonia couldn't. I'm experimenting with it to see if this can be fixed. Still, being able to set the duration for a sound seems useful.

nc7s:
The culprit of rodio being unable to read duration should be https://github.com/RustAudio/rodio/issues/696, implemented in https://github.com/RustAudio/rodio/pull/697, merged early last month. I've asked if they could release it to avoid a git+rev dependency.

nc7s:
FWIW, if we are to not disturb the user listening to some music, `duration.unwrap_or(Duration::MAX)` would allow the thing to be played to practically the infinity. We could even use this to get rid of the `duration` field of embedded sounds.

c-git:
I wouldn't over index on it. I might be the only person who sets long running tasks in bacon and wants a longer sound.

Canop:
@nc7s I made a lot of comments, but I now see that they're "pending". 

image

Those comments never reached you, right ? ^^

Canop:
If you confirm that I messed up sending the review, I'll redo it.

nc7s:
> @nc7s I made a lot of comments, but I now see that they're "pending".
>
> (... screenshot)
> 
> Those comments never reached you, right ? ^^

Nope, they are still pending.

> If you confirm that I messed up sending the review, I'll redo it.

I confirm that GitHub's UI isn't the clearest about whether a review session has been committed :> No need to redo I think, just submit. Ignore me if it's already gone.

nc7s:
Well, maybe you do need to redo it..

screenshot showing Canop has 'self-requested' a review

Canop:
There are problems with the PR as is:
- an unknown duration makes the player thread wait until bacon quits, and prevents following sounds
- the sound is loaded (which can error) and *leaked* in SoundConfig. This object is often created many times on job creations (on each layer of config merging) so it's a real problem not just a one time leak. The current gard is also a bug as it makes an earlier config win over a later one (opposite of the overall design)
- importing a heavy crate just to handle tildes in paths (there's already directories-next in the tree)
- only MP3 are supported today

Duration:
You inherited the structure I had devised initially for a fixed set of sounds where having predetermined durations wasn't a problem.
This duration was meant to cut the sounds because the files were often much longer than what can be heard in them and it prevented repeated sounds and delayed quit.
But most user-provided sounds would probably not have the problem so this duration can become optional.
When no duration is known, we can end when sink.empty().

Loading:
The SoundPlayer should be the one loading the file. This would happen only once per mission and wouldn't leak.

I suggest I take from there and do the changes, if you don't mind, so that you don't have to battle trying to apply my design choice. Are you OK with this?

nc7s:
Other than a bit sad I didn't get to make it work better with the existing code and the general considerations behind it, sure, I'm glad you would take this into consideration ;)

Canop:
The feature is in main: https://dystroy.org/bacon/config/#sound

PR #354, opened 2025-05-05
add support for nextest 0.9.95

### LINKED ISSUE #350: Upcoming changes to nextest output
Hi!

Just wanted to let you know of some upcoming changes to nextest's output for failed tests:

* the stdout and stderr headers have become less visually busy
* there's now a 4-space indent

See https://nexte.st/docs/features/leaky-tests/#marking-leaky-tests-as-failures for an example.

These changes are in cargo-nextest 0.9.95-rc.2, and I'd love it if you updated bacon to be compatible! Hopefully these will be the last set of output changes for a while.

(BTW there are also some correctness fixes to [libtest JSON output](https://nexte.st/docs/machine-readable/libtest-json/) in 0.9.95. I'd definitely want to have a stable machine-readable format at some point, but it's a major endeavor as you might imagine.)

Canop:
Fix #350

### DIFF
--- a/src/analysis/nextest/nextest_line_analyser.rs
+++ b/src/analysis/nextest/nextest_line_analyser.rs
@@ -90,9 +90,12 @@ impl NextestLineAnalyzer {
         }
         extract_key_after_crate_name_v2(strings)
     }
-    /// Return the last test key when the line is like "──── std(out|err)"
-    /// For nextest 0.9.95+ (see https://github.com/Canop/bacon/issues/350)
+    /// Return the last test key when the line is like either
+    ///  - "──── stderr"
+    ///  - "── stderr ──" (with --no-output-indent)
+    /// (with stderr maybe replaced by stdout or output)
     ///
+    /// This supports nextest 0.9.95+ (see https://github.com/Canop/bacon/issues/350)
     fn stdx_section_key_v3(
         &self,
         content: &TLine,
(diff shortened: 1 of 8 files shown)


src/analysis/nextest/nextest_line_analyser.rs
```diff
@@ -48,31 +53,68 @@ impl LineAnalyzer for NextestLineAnalyzer {
...
+    ///
+    fn stdx_section_key_v3(
+        &self,
+        content: &TLine,
+    ) -> Option<String> {
+        let mut strings = content.strings.iter().skip_while(|ts| ts.is_blank());
+        let ts = strings.next()?;
+        if ts.csi != CSI_ERROR || (ts.raw != "stderr" && ts.raw != "stdout") {
+            return None;
+        }
```
sunshowers, line 104: While you're here I'd also support `output`, which is for situations where stdout and stderr are combined. This is currently semi-hidden but might become more visible soon.

src/analysis/nextest/nextest_line_analyser.rs
```diff
@@ -48,31 +53,68 @@ impl LineAnalyzer for NextestLineAnalyzer {
...
+        if !ts.is_blank() {
+            return None;
+        }
+        let ts = strings.next()?;
+        if ts.csi != CSI_ERROR || ts.raw != "───" {
+            return None;
+        }
+        if strings.next().is_some() {
+            return None;
+        }
```
sunshowers, line 115: The output is indented by 4 spaces now -- do you want to unindent it?

sunshowers (commented):
Thanks!

src/analysis/nextest/nextest_line_analyser.rs
Canop: Wouldn't it be better to keep the indentation, and let users who don't want indentation just pass `--no-output-indent` in the nextest command ?

sunshowers: Sure, depends on the UX you'd like to present to users. (Wasn't sure if you do your own formatting that might interfere with the indent.)

The `stdout` and `stderr` lines are slightly different with `--no-output-indent` -- would be worth supporting those too. (I tried to keep them the same, but the slight difference just looks better.)

image

Canop: Should be OK now

src/analysis/nextest/nextest_line_analyser.rs
Canop: Added "output"

PR #355, opened 2025-05-05
bacon listen for actions on a unix socket

### LINKED ISSUE #352: Connect to editor and adapt presentation
Bacon is great, but it would be even more useful if it could connect to Vim (or Vs Code ?);

* Sort errors with currently opened file first
* Run only test for current file or even single test where cursor is

And maybe
* Select error to jump to file & line
Although it goes against the non-interactive idea, and frankly not as useful as having bacon just adapt its output according to your location in the editor.

Canop:
Either add `listen = true` in bacon configuration or launch bacon with `--listen`.

Then, it's possible to send actions on the unix socket `.bacon.socket` in the package directory.

A simple way to try it is to launch `socat - UNIX-CONNECT:bacon.socket` then issue actions ending in new lines.

image

You can also use bacon to send just one command, eg `bacon --send "job:clippy"`

This new `listen` mode comes with a new action: `focus-file` which tells bacon to show the diagnostics related to a specific file first. It might be made more powerful later (ranges, several files, etc).

Example:

     bacon --send 'focus-file(cmd_result.rs)'

(it works with a partial path, end of the path is tested)

Fix #352


sasq64:
This works fine I think!
I would prefix the socket file with a dot to hide it though.

And then perhaps add a `focus <file> [range]` command, telling bacon to first show messages from `file`, and if given, first show errors with lines in `range` ?

Then a vim plugin could by default call focus with the file and visible lines when they change

Canop:
Added `bacon --send`

>  I would prefix the socket file with a dot to hide it though.

Done

> And then perhaps add a focus <file> [range] command, telling bacon to first show messages from file, and if given, first show errors with lines in range ?

I'll think about it.

sasq64:
I tried to make this command work but couldn't figure out what syntax it accepts :)

Canop:
> I tried to make this command work but couldn't figure out what syntax it accepts :)

Sorry. I've added an example in the first post.

PR #365, opened 2025-06-04
feat: use polling to watch files

### LINKED ISSUE #313: File watcher doesn't work on mounted files on Docker (Windows host)
I am running `bacon --headless run-long` on a docker dev container to hot reload. Container is using volume mounted from Windows host and bacon should be watching the files on that volume. Running on windows and running on linux works fine. But for some reason file watcher doesn't seem to work on this specific case.

I think bacon checks for the file changes by listening inotify events. Is there any option that uses polling instead that I have missed?

Odonno:
Fix #313

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1727,11 +1727,11 @@ dependencies = [
 
 [[package]]
 name = "inotify"
-version = "0.10.2"
+version = "0.11.0"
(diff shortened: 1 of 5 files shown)


Canop:
I'd like feedback from users on this one: Does it make bacon run better in some context (docker maybe) ?

(I can take care of the conflicts if it's proven to solve problems)

PR #397, opened 2025-10-22
Config schema

### LINKED ISSUE #372: Add a jsonschema for config format to make config autocomplete
It would be nice to have auto completion of the config properties when edting the toml file.
Many toml editors will automatically load the schema from schemastore.org.

See https://github.com/release-plz/release-plz/blob/main/crates/release_plz/src/generate_schema.rs for some code on how to automate generating this (release-plz generates the schema for its config file)

Canop:
Generate a JSON schema for the configuration files

Fix #372 

Usage:

    bacon --generate-config-schema > bacon.json

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -150,7 +150,7 @@ pub struct Args {
     pub config_toml: Option<String>,
 
     /// Generate the JSON Schema for bacon configuration files
-    #[clap(long)]
+    #[clap(long, hide = true)]
     pub generate_config_schema: bool,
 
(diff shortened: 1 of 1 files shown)


Canop:
@joshka Self-hosting looks more reliable than hosting at schemastore.org: I can have the update of the schema be part of the release deployment script. 

So I'll have this directive at the top of default config files:

`#:schema https://dystroy.org/bacon/.bacon.schema.json`

As my tools usually accept previous versions of configuration files, I won't add a version suffix unless it really becomes necessary.

joshka:
Nice! Thanks for pushing that through.

PR #398, opened 2025-11-02
Update config docs

joshka:
Updates #397 with extra schema docs.

Generated by codex, validated by hand as accurate to the best of my understanding of the options documented.

Tested by experimenting how VSCode displays the descriptions on a few different items in bacon.toml (using [Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml) extension) - see https://taplo.tamasfe.dev/configuration/using-schemas.html#using-schemas for config approaches. The description of "command" under each job doesn't seem to pick up the description. I'm not sure why - possibly some sort of problem in the extension. The schema seems valid.

### DIFF
--- a/bacon.json
+++ b/bacon.json
@@ -251,7 +251,7 @@
       "$ref": "#/$defs/BaconSkin"
     },
     "sound": {
-      "description": "Per-job sound settings that override the global configuration.",
+      "description": "Notification sounds",
       "$ref": "#/$defs/SoundConfig"
     },
     "summary": {
@@ -970,7 +970,7 @@
       }
     },
     "Exporter": {
-      "description": "Export backend available for Bacon exports.",
+      "description": "Export backend.",
       "oneOf": [
         {
           "description": "The analyzer is tasked with doing an export while analyzing the\ncommand output",
@@ -1162,7 +1162,7 @@
           "$ref": "#/$defs/BaconSkin"
         },
         "sound": {
-          "description": "Per-job sound settings that override the global configuration.",
+          "description": "Notification sounds",
           "$ref": "#/$defs/SoundConfig"
         },
         "watch": {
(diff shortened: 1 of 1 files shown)


src/conf/config.rs
```diff
@@ -18,41 +18,52 @@ use {
...
+    /// Custom keybindings layered on top of the defaults.
     pub keybindings: Option<KeyBindings>,
 
+    /// Whether to display the mission output in reverse order.
     pub reverse: Option<bool>,
 
+    /// Whether to show the condensed summary area by default.
     pub summary: Option<bool>,
 
+    /// Deprecated toggle that enables a built-in set of Vim-style keybindings. Use `keybindings` instead.
```
joshka: While the deprecated tag causes `"deprecated": true` to be added to the schema, not all places understand this, and the description (use keybindings) doesn't flow through to the UI, so it's worth keeping this info here regardless.

joshka (commented):
Some self-review comments

src/jobs/job.rs
```diff
@@ -99,11 +100,14 @@ pub struct Job {
...
     pub on_success: Option<Action>,
 
+    /// Whether to display how many files triggered the current run.
     pub show_changes_count: Option<bool>,
 
+    /// Color overrides applied to the UI while the job runs.
     #[serde(default)]
     pub skin: BaconSkin,
 
+    /// Per-job sound settings that override the global configuration.
```
Canop: Job fields are flattened into the global config too, so you can't say it's "per-job".

joshka:
```suggestion
    /// Notification sounds
```

src/sound/sound_config.rs
```diff
@@ -4,9 +4,13 @@ use {
     serde::Deserialize,
 };
 
+/// Configuration for mission sound notifications.
```
Canop: Let's keep it simple: "Sound configuration.

Canop:
BTW, thanks for pointing the Even Better TOML extension, which helps me test the schema.

Canop (changes requested):
Looks good to me, just a few minor suggestions

PR #400, opened 2025-11-11
Support more 4-bit colours in `rustc` analyser

narpfel:
Corresponding `rustc` change: https://github.com/rust-lang/rust/pull/147207

Resolves #399.

PR #402, opened 2025-11-12
Fix Windows build after #400

narpfel:
`CSI_BOLD_4BIT_BLUE` is used outside of `#[cfg(not(windows))]`, so it should not be `cfg`’d out on Windows.


Canop:
Already fixed.

Issue #403, opened 2025-11-14
`bacon` sometimes misses non-zero exit codes


narpfel:
Usually when a command run through `bacon` exits with a non-zero exit code, `bacon` prints `Command error code: <code>` in the top status bar. This does not always work.

Small reproducer (in an empty `cargo` project):
`src/main.rs`:
```rust
use std::process::ExitCode;

fn main() -> ExitCode {
    println!("program ran");
    ExitCode::FAILURE
}
```

No `bacon.toml`/global `bacon` config file.

Run `bacon run` and keep pressing `r`. Eventually there will be a run where the `Command error code` message is not shown.

Alternatively, this small script runs `bacon` in a loop until the message is missing:
```bash
while true; do
    (sleep 1s; killall bacon) & /usr/bin/bacon run > out 2>&1
    if grep -q "program ran" out && ! grep -q "Command error code: 1" out; then
        break
    fi
done
```

For me, it took about a minute to reproduce with this script.

For this simple program, it doesn’t happen very often (maybe 1% of the time), but in a real project (which prints about 120 lines and overall takes about 200ms to run), this happens at a noticeable rate, maybe 50%.

Slightly larger reproducer that makes manual reproduction easier:
```rust
use std::process::ExitCode;
use std::time::Duration;

fn main() -> ExitCode {
    for i in 0..1000 {
        println!("{}", i.to_string().repeat(i));
    }
    std::thread::sleep(Duration::from_millis(150));
    println!("program ran");
    ExitCode::FAILURE
}
```

More output and longer execution times seem to make it happen more often.

As a small check that it’s not `cargo` that’s swallowing the exit code, this does not terminate for me:
```bash
while ! cargo run; do true; done
```

narpfel:
I think I found the problem: When the child process closes its `stderr`, `bacon` assumes that the process has ended https://github.com/Canop/bacon/blob/736cd69ca440bdd11ecec2f7cfb37f3400cee66d/src/exec/executor.rs#L178-L183 and tries to read the child’s exit code using `try_wait`: https://github.com/Canop/bacon/blob/736cd69ca440bdd11ecec2f7cfb37f3400cee66d/src/exec/executor.rs#L202

[...]

PR #405, opened 2025-11-24
Fix race condition between child process closing `stderr` and reading the exit code

narpfel:
This fixes the issue reported in #403, but only for processes that close `stderr` as part of their shutdown. When the child process closes `stderr` early, this will still cause `bacon` to freeze (see #403 for a small reproduction), but it will no longer ignore all further output on `stdout`.


Canop:
Note: I didn't miss this PR. I just have too much professional work ATM so it might take a few weeks before I can review it.

narpfel:
No worries!

PR #409, opened 2025-12-02
Fixup MSRV snafu, allow developers to use new resolver without forcing edition2024

### LINKED ISSUE #407: Cargo.lock should be compatible with MSRV
This project's [Cargo.toml](https://github.com/Canop/bacon/blob/v3.20.1/Cargo.toml#L12) currently claims it supports Rust >= 1.77

But with Rust 1.87, `cargo install --locked bacon` fails with:

```
Caused by:
  rustc 1.87.0 is not supported by the following package:
    home@0.5.12 requires rustc 1.88
```

alerque:
Closes #407, see also discussion [in chat](https://miaou.dystroy.org/4683?bacon#18104413).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,18 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "ahash"
-version = "0.8.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


alerque:
(Draft mode because there are still a couple deps that don't match the MSRV, rooting out necessary downgrades is harder than just using the right resolver to not bump the lock file to too new in the first place!).

.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+
+[resolver]
+incompatible-rust-versions = "fallback"
```
Canop: Oh... I didn't know this option...

alerque:
Please don't squash these commits on merge, they are carefully split up for a reason ;-)

PR #424, opened 2026-01-07
Add {job} and {item_idx} templates, add select-item(idx) action

### LINKED ISSUE #423: Add {job} and {item_idx} to locations export and select-item(n) socket action
While working on the emacs minor mode (#422), I'd like to propose features that would enable better editor integration:

1. Add {job} and {item_idx} variables to locations export line_format
Currently the locations export supports: {kind}, {path}, {line}, {column}, {message}, {context}

Adding {job} (from mission.concrete_job_ref.badge_label()) and {item_idx} (from line.item_idx) would allow:

```
[exports.locations]
line_format = "{kind}@{job}[{item_idx}] {path}:{line}:{column} {message}"
```

This would output:

```
error@clippy[1] /home/user/src/foo.rs:42:5 mismatched typeswarning@clippy[2] /home/user/src/bar.rs:10:1 unused variable
error@test[3] /home/user/src/baz.rs:100:1 test failed
```

Both values are already available in the export code path:

job - via mission.concrete_job_ref.badge_label() (Mission is already passed to write_locations)
item_idx - via line.item_idx (already tracked on every Line struct)

Use case: Editor plugins can:

Know which job produced each diagnostic (useful when switching jobs)
Use the item index for sync back to bacon when jumping from error to error.

2. Add select-item(n) socket action

A select-item(n) action would scroll bacon's display to show item n at the top (similar to typing :n in bacon's goto mode).

Use case: When a user navigates to error [3] in their editor, the editor plugin can send select-item(3) to bacon, keeping both views in sync.

This builds on the existing socket infrastructure from PR #355 and would make editor integrations much more cohesive.

Note: {job} is unused in this example, eventually it may become useful, the editor mode can cache locations and switch jobs then

cehteh:
This closes #423 and implements the proposed things.
5a36b6 also adds some more documentation about what template variables are available and gives a short example.


website/src/config.md
```diff
@@ -197,6 +197,7 @@ scroll-pages(-1) | <kbd>PageUp</kbd> | move one page up
 scroll-pages(1) | <kbd>PageDown</kbd> | move one page down
 scroll-to-bottom | <kbd>End</kbd> | scroll to bottom
 scroll-to-top | <kbd>Home</kbd> | scroll to top
+select-item(n) | | scroll to display the diagnostic item with index n at the top
```
Canop: As there's no real selection but just a scrolling, the name might be a problem in the future. `scroll-to-item` or `show-item` look better.

cehteh:
On Wed, 07 Jan 2026 12:44:43 -0800
Denys Séguret ***@***.***> wrote:

> @Canop commented on this pull request.
> 
> 
> 
> > @@ -197,6 +197,7 @@ scroll-pages(-1) | <kbd>PageUp</kbd> | move one
> > page up  
>  scroll-pages(1) | <kbd>PageDown</kbd> | move one page down
>  scroll-to-bottom | <kbd>End</kbd> | scroll to bottom
>  scroll-to-top | <kbd>Home</kbd> | scroll to top
> +select-item(n) | | scroll to display the diagnostic item with index
> n at the top
> 
> As there's no real selection but just a scrolling, the name might be
> a problem in the future. `scroll-to-item` or `show-item` look better.
> 

your call/decision, I was also thinking about just goto(idx), if I
have to choose maybe show-item() is favorable. Brainfart was goto(job,
index) .. but since job switching is rather expensive and indices may
be not up to date, this may be not viable.

Canop:
Another cosmetic detail: kebab-case is favored in bacon configuration over snake_case, so it should probably be `{item-idx}` for consistency.

Canop:
This works well in my tests.

Canop:
Merged, thanks

PR #425, opened 2026-01-08
feat: support negative `!pattern` syntax in ignore to override .gitignore

### LINKED ISSUE #308: Don't ignore files that have been added to a job's watch list
I'm doing some interactive testing on a temp file that I don't want to commit, using a file name that's in my global `.gitignore`.

So, I have a job where I've added the file to the `watch` list, and then I'd like it if `GitIgnorer` would make an exception for the explicitly watched file, instead of telling the watcher that it's OK to ignore it.

ishad0w:
Adds support for `!pattern` syntax in the `ignore` config field to override gitignore rules.

```toml
watch = ["."]
ignore = ["!.env"]
```

Fixes #308.

### DIFF
--- a/src/ignorer/glob_ignorer.rs
+++ b/src/ignorer/glob_ignorer.rs
@@ -1,4 +1,5 @@
 use {
+    super::build_glob_patterns,
     crate::*,
     anyhow::Result,
     std::path::Path,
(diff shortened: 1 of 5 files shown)


Canop:
Ok, this looks good but I'll need to test a little and rethink about it before I merge it and I'd welcome other user feedback.

PR #427, opened 2026-01-11
Analyzers for Typescript and Go compilers

faultymajority:
Quick attempt at extracting Typescript type checking errors.

One way to drive this in an existing Typescript project:
```
[jobs.tscheck]
command = [
    "tsc",
    "--noEmit", "--pretty",
]
need_stdout = true
analyzer = "typescript"
watch = ["src"]
```

### DIFF
--- a/website/src/analyzers.md
+++ b/website/src/analyzers.md
@@ -9,17 +9,19 @@ This page is an overview of the supported tools and how bacon can be configured 
 Analyzer | Languages | Tool
 -|-|-
 [standard](#rust) (*default*) | Rust | [cargo](https://doc.rust-lang.org/cargo/) `check`, `build`, `test`, `clippy`, `doc`, `run`, `miri`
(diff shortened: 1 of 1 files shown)


src/analysis/typescript.rs
```diff
@@ -0,0 +1,97 @@
...
+    if let Some((_, location, code)) = regex_captures!(r"([^\s:]+:\d+:\d+) - error (TS.+$)", &raw) {
+        return Some(LocationCode {
+            location: location.to_string(),
+            code: code.to_string(),
+        });
+    }
+    None
+}
+
+/// Build a report from the output of biome
```
Canop: "biome" ? Is that due to a hasty copy-pasting ?

faultymajority:
Added an analyzer for Go as well. Example config:

```
[jobs.gocheck]
command = [
    "go", "build",
    "-v",
    "./...",
]
need_stdout = true
analyzer = "go"
watch = ["."]
```

PR #430, opened 2026-01-17
make radicle id visible

cehteh:
Doh, was a bit fast earlier. Due to the distributed nature web-frontend in radicle are rather advisory.


Canop:
I had modified before, so I just copied this new sentence.

Issue #438, opened 2026-03-18
Way to debug what triggered a re-run


marcospb19:
I have the following job config:

```
[jobs.test]
command = ["cargo", "test"]
ignore = ["tests/snapshots"]
```

And I'm running `bacon test`.

My tests modify the folder "tests/snapshots" ([this lib](http://docs.rs/insta)), which causes tests to trigger runs in a loop.

So I added the `ignore = ["tests/snapshots"]` line, the problem is, it's not working, still running in loop.

`cargo-watch` has a `--why` flag that tells why it's re-triggering, it looks like this:

```
2026-03-18T18:36:39.461-03:00 - INFO Paths updated: [PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_err_decompress_output_file_exists.snap.new", op: Some(WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_err_decompress_output_file_exists.snap.new", op: Some(CLOSE_WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_ok_decompress_multiple_files.snap.new", op: Some(WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_ok_decompress.snap.new", op: Some(WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_decompress_with_unknown_extension_shows_output_path.snap.new", op: Some(WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_decompress_with_unknown_extension_shows_output_path.snap.new", op: Some(CLOSE_WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_ok_decompress_multiple_files.snap.new", op: Some(CLOSE_WRITE), cookie: None }, PathOp { path: "/home/marcospb19/ouch/tests/snapshots/ui__ui_test_ok_decompress.snap.new", op: Some(CLOSE_WRITE), cookie: None }]
```

Here's that but readable:

```
/home/marcospb19/ouch/tests/snapshots/ui__ui_test_err_decompress_output_file_exists.snap.new
/home/marcospb19/ouch/tests/snapshots/ui__ui_test_err_decompress_output_file_exists.snap.new
[...]

PR #439, opened 2026-03-18
display which file triggered the re-run

marcospb19:
A draft to address #438.

What it looks like (at the time of writing):
image

TODO: 
- [ ] Pressing F5 wrongly shows the last triggered path instead, instead of "computing..."

### DIFF
--- a/src/tui/mission_state.rs
+++ b/src/tui/mission_state.rs
@@ -4,6 +4,7 @@ use {
     crokey::KeyCombination,
     std::{
         io::Write,
+        path::PathBuf,
         process::ExitStatus,
         time::Instant,
     },
@@ -79,6 +80,8 @@ pub struct MissionState<'a, 'm> {
     pub search: SearchState,
     /// The dialog that may be displayed over the rest of the UI
     pub dialog: Dialog,
+    /// The last seen path that triggered a re-run.
+    pub last_trigger_path: Option<PathBuf>,
 }
 
 impl<'a, 'm> MissionState<'a, 'm> {
@@ -134,6 +137,7 @@ impl<'a, 'm> MissionState<'a, 'm> {
             dialog: Dialog::None,
             app_state,
             mission,
+            last_trigger_path: None,
         })
     }
     pub fn open_jobs_menu(&mut self) {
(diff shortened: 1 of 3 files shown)


src/tui/mission_state.rs
```diff
@@ -826,12 +834,33 @@ impl<'a, 'm> MissionState<'a, 'm> {
...
         let width = self.width as usize;
         if self.computing {
+            let text = if let Some(last_trigger_path) = &self.last_trigger_path {
+                let path_display = last_trigger_path.display().to_string();
+
+                let prefix_len = "watch trigger: ".len();
+                let padding_len = 2;
+
+                let should_be_shortened =
+                    last_trigger_path.as_os_str().len() + prefix_len + padding_len > width;
```
Canop, line 844: counting bytes isn't the right way to know if a string should be shortened. Even counting chars is wrong, as some may be double width. And you can't shorten a string at any arbitrary byte, or the UTF8 might be invalid.

But there are proper utilities in bacon do display strings while taking care of the available size. Look for calls to `write_composite_fill` in bacon (the last argument can be `Center`).

Canop:
I don't think this is the right approach to debug bacon. You may have much more to display and you can't always write this info in bacon itself.

You should use the log: launch bacon with `BACON_LOG=debug bacon` (or  `BACON_LOG=debug cargo run -- path/to/project` if you're in the bacon project dir. Then display the log with `tail -f bacon.log` in another terminal. You'll see for example that the path triggering the recompute is logged, as well as some ignoring related info. And you can of course, to find the problem, add more `debug!` and `info!` in the code.

marcospb19:
> You should use the log: ...

Sounds great, thanks! I couldn't find that in the docs, so I'll pass my eyes on it again later to see if I just missed it, or if I can make a PR to improve it.

Canop:
You're right, it's missing in the doc.

It should be in this page: https://dystroy.org/bacon/community/bacon-dev/

And it should be similar to what I've written for another program: https://dystroy.org/broot/community/#log

Canop:
Actually, logging is mentioned: http://localhost:8004/community/bacon-dev/#issues

marcospb19:
Hey @Canop, sorry for the delay.

Before I proceed, are you still open to accepting this PR and do you have any nitpicks for how it's displayed in the screenshot?

Instead of showing the full path, I'd prefer having it display the relative path (when it's inside of the current folder), but then, the code that displays need to either read the current folder every time it's rendered, or store it in the struct, so not the prettiest code, any thoughts?

Canop:
Usefulness can be debatted: There are often several files involved in a change, and you can always display them in the log.

If it's cleanly done and gated by a configuration setting, this could still be interesting to some users, though, so I'm open.

I think it's preferable to have the whole path, but to use termimad's ability for internal elision. See the [fit-width example of termimad](https://github.com/Canop/termimad/tree/main/examples/fit-width):

image

marcospb19:
I think I'll stick to the BACON_LOG then, as it is intended, thanks for the troubleshooting help tho!

PR #441, opened 2026-05-10
enable tracing in rodio to stop use of eprintln

### LINKED ISSUE #437: "New" message after audio
I missed a few version of bacon and upgraded today

```sh
    Replaced package `bacon v3.16.0` with `bacon v3.22.0` (executable `bacon`)
```

I upgraded because I was getting an "issue" where the output of the bacon run would disappear and show failed waiting for other jobs to finish. So updated and the issue is no longer there but now I get a message for a short while at the bottom of the screen (what seems like every time) when audio is played. It doesn't really cause me any problems and I'm fine to ignore it but wondered if this was intended / known. So mentioning it.

The message says "Dropping OutputStream, audio playing through this stream will stop" (See screenshot below). Fine to close this issue if this is intended just wasn't sure if it was indeed intended, I'm not sure if as many users use audio hence I decided to mention it.

Image

c-git:
Fixes the message that shows after sounds stop

resolves #437

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2340,6 +2340,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
+name = "pin-project-lite"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
+
+[[package]]
 name = "pkg-config"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Canop:
Just to be sure: when the "tracing" feature is enabled, rodio logs instead of printing when it's not happy with the stream being dropped before its end ?

c-git:
Yes that's correct. I included a link to the part of the code that led me to believe that. I don't think the message is identical in the version we are using but the logic is the same. I Ctrl clicked to get to the implementation so RA took me to the version we are using which doesn't mention the other option. But the logic is the same. I also tested it after and I confirmed that I no longer see the message.

https://github.com/RustAudio/rodio/blob/master/src%2Fstream.rs#L103

c-git:
Back on the computer now here's the actual code from the version we use

```rust
impl Drop for OutputStream {
    fn drop(&mut self) {
        if self.log_on_drop && !std::thread::panicking() {
            #[cfg(feature = "tracing")]
            tracing::debug!("Dropping OutputStream, audio playing through this stream will stop");
            #[cfg(not(feature = "tracing"))]
            eprintln!("Dropping OutputStream, audio playing through this stream will stop")
        }
    }
}
```

PR #442, opened 2026-05-12
scroll_anchor setting

### LINKED ISSUE #384: [feature] Add option to tail the output/stick to scrolling to bottom
Currently, if there's not enough space in the terminal to show everything, bacon will show the top part of the output and hide the latest/tail output. This is fine for the common use-cases like `bacon check`, where the first error is often what one wants to see, but for `bacon run` it might be suboptimal. It would be great to have an option to always see the latest/tail output of the running command instead.

I know there's `reverse`, but it's not the same as it puts everything upside down.

So it would be great to have an option, maybe just `tail (true/false)` or similar that by default is disabled, but when enabled makes bacon always tail the output. Basically a sticky version of the already implemented `scroll-to-bottom`.

Happy to help with the implementation.

Canop:
bacon sorts items, with errors first, followed by warnings.

The first items (on top if you don't have the *reverse* setting) are usually the most interesting ones.

But not always. For example, with the `run` job, when there's no error and the compiled binary is effectively executed, you may want to focus on the output of the program.
And if it's a `long run`, you probably want to have the scroll follow the new lines unless you decided to scroll away.

This PR introduces the `scroll_anchor` setting, whose value can be undefined, `auto`, `first`, or `last`.

When not defined, the applied default will be `first`.

In the default `bacon.toml`, the `run` and `run-long` jobs have `scroll_anchor = "auto"` which means that:

* if errors were recorded, the default sticky position is the first item
* if there was no error, the default stiky item is the last one (bacon then acting as `tail` to follow new lines)

This should fix #384 but there will be a test period before I try to merge it, or not. Feedback welcome.

### DIFF
--- a/defaults/default-bacon.toml
+++ b/defaults/default-bacon.toml
@@ -87,6 +87,7 @@ need_stdout = true
 allow_warnings = true
 background = true
 show_command_error_code = true
+scroll_anchor = "auto"
 
 # Run your long-running application (eg server) and have the result displayed in bacon.
 # For programs that never stop (eg a server), `background` is set to false
@@ -108,6 +109,7 @@ allow_warnings = true
 background = false
 on_change_strategy = "kill_then_restart"
 # kill = ["pkill", "-TERM", "-P"]
+scroll_anchor = "auto"
 
 # This parameterized job runs the example of your choice, as soon
 # as the code compiles.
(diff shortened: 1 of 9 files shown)


c-git:
Yeah this makes sense to me. I'll switch to this version and let you know what I think. If you want feedback before I feel I have anything useful just ping me and I'll tell you want I think so far. I don't quite have time to specifically test this but I will switch to this version and let you know how I feel using it. I've overridden the default long run job so I'd need to add the setting to auto to my job right?

Canop:
> I'd need to add the setting to auto to my job right?

Yes. All user defined jobs are on "unspecified" which means "first".

c-git:
I added `scroll_anchor = "auto"` to the top section of my `prefs.tom` because I can't at this point see when I wouldn't want that. Will let you know how it goes.

PR #444, opened 2026-06-22
Use the level provided in json for cargo_json analyzer

hkalbasi:
Hi! First, thanks for the bacon, it is a very nice tool.

I'm working on a [new language CO2](https://github.com/hkalbasi/co2) which is basically C with some Rust interop features. It uses a cargo wrapper `co2cargo` which tries to be highly compatible with cargo.

I tried to use bacon with `co2cargo`, and it failed, because I use a slightly different formatting for diagnostics. I found out that bacon has a `cargo_json` format analyzer, and `co2cargo` supports cargo's json format. Problem solved, right?

It turns out that `cargo_json` is not very different from the text based analyzer, and just run the text based analyzer on the `rendered` field of the json. So it didn't help in my case, and I decided to make this PR to address that.

This PR slightly improves the situation, by obtaining the level of the diagnostic from the json `level` field, not parsing it from the `rendered` field. This makes `cargo_json` format more robust from future changes in the text format, and enables third party tools (like `co2cargo`) who emit diagnostics compatible with `cargo_json` but not using the same text format to be usable with bacon.

IIUC, current bacon design is hostile to structured formats, and they are second class compared to text based parsers. Which is sad because text based formats are not stable in most tools, and will eventually break, specially now that bacon is expanding its analyzers to support more tools from more language. Part of the problem is that bacon does its ui elements on top of the text emitted by the tool, but I still think there is room for improvements on structured formats handling.

### DIFF
--- a/src/analysis/cargo_json/mod.rs
+++ b/src/analysis/cargo_json/mod.rs
@@ -3,14 +3,18 @@ mod cargo_json_export;
 use {
     super::*,
     crate::{
-        analysis::standard,
+        analysis::standard::build_report_from_analysis,
         *,
     },
     anyhow::Result,
+    lazy_regex::*,
     cargo_json_export::*,
     cargo_metadata::{
         Message,
-        diagnostic::Diagnostic,
+        diagnostic::{
+            Diagnostic,
+            DiagnosticLevel,
+        },
     },
 };
 
@@ -26,7 +30,7 @@ use {
 /// not sure this is worth it.
 #[derive(Default)]
 pub struct CargoJsonAnalyzer {
-    lines: Vec<CommandOutputLine>,
+    analysis: Vec<(LineAnalysis, TLine)>,
     exports: Vec<CargoJsonExport>,
 }
 
@@ -35,7 +39,7 @@ impl Analyzer for CargoJsonAnalyzer {
         &mut self,
         mission: &Mission,
     ) {
-        self.lines.clear();
+        self.analysis.clear();
         self.exports.clear();
         for (name, export_settings) in &mission.settings.exports.exports {
             if export_settings.exporter == Exporter::Analyser {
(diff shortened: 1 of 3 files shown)


Canop:
> there is room for improvements on structured formats handling

Yes. The reason this work stopped is that there was no good reason to go that way. I mostly implemented the cargo_json analyzer as a backup solution in case the text output of cargo was too badly changed to be usable.

Textual format has a big upside: it's easily read by humans. IMO the best solution would be a cleaner, generic enough, textual format, ideally a superset of cargo's output but only if it's feasible. I mentioned it here: https://dystroy.org/blog/bacon-everything-roadmap/#introduce-burp

hkalbasi:
I think if the `Report` structure was a `Vec<Item>` where item was:
```Rust
struct Item {
    kind: Error | Warning | TestFail | ...,
    location: FilePath,
    // Any other metadata needed for bacon?
    rendered: AnsiColoredString,
}
```
Instead of the current annotated lines structure, it would make it much easier to write analyzer for structured outputs.

> Textual format has a big upside: it's easily read by humans.

How it is relevant? The data transferred between bacon and the tool is not visible to humans. The upside for structured output is that they won't break, and they are easier to parse for analyzer code.

I can try to experiment with that `Vec<Item>` as a refactor with no visible user change, if you want.

Canop:
> How it is relevant? The data transferred between bacon and the tool is not visible to humans. 

The tool must, in both cases, transfers the same information. If you put that in two separate formats, there's always a risk of them not having exactly the same content. And when this promise breaks, then you don't fix it because "the structured output didn't break" which means you're not showing what you should to the user. You're making a tool for a tool, not for the human.

> I can try to experiment with that Vec<Item> as a refactor with no visible user change, if you want.

There's already a project in (slow) progress regarding a new structure of the report. So I'd rather not have new centralized experimentation added (but a separate analyzer managing its own format as intermediate structure is possible).

PR #445, opened 2026-07-07
Add a clear-output action

### LINKED ISSUE #419: I want a simple shortcut key to clear the screen. How can I set it up?

Canop:
This can be tested by adding a key binding such as

`x = "clear-output"`

Fix #419


bachirelkhoury:
works well thanks @Canop!

PR #453, opened 2026-07-25
Improve executor on-kill behavior

Canop:
With the (not default) kill-then-restart onchange strategy, the previous code could result in some lines leaking from the killed task to the following one (race condition, never observed)


Canop:
I now think this approach is too dangerous, as it can probably freeze bacon on cargo failing to kill its own children (due to the added `join`).

Issue #205, opened 2024-08-13
Hard coded build script means ineffective watches when the file is in a non standard location


alerque:
The location of `build.rs` to watch seems to be hard coded as the Cargo default:

https://github.com/Canop/bacon/blob/27060c83353c9bdfc4f2aebad3032f7105d5e67e/src/mission.rs#L17

This location is, however, configurable. I do so in most of my projects to group build tooling out of the way of the top level directory when possible:

```toml
[package]
build = "build-aux/build.rs"
```

This location is not watched, hence changes to the file do not trigger updates in `bacon`.

I would suggest the Cargo manifest actually be read and the value of `build` (if set), and fall back to the default if it is not set. There might even be a way to query the active value out of Cargo, but I don't know what it is. Somewhat unhelpfully `cargo read-manifest` buries the relevant info a bit. For example on my [git-warp-time](https://github.com/alerque/git-warp-time) project that relocates the build script, the info ends up here:

```console
$ cargo read-manifest | jq -r '.targets[] | select(.kind[0] == "custom-build") | .src_path'
/home/caleb/projects/git-warp-time/build-aux/build.rs
````

Canop:
> I would suggest the Cargo manifest actually be read and the value of build (if set)

I'll have a look.

In the meantime, do we agree that you can simply add `watch = ["build-aux/build.rs"]` to your jobs ?

alerque:
> In the meantime, do we agree that you can simply add `watch = ["build-aux/build.rs"]` to your jobs?

I haven't tried it yet, but possibly yes. Is that value additive? Or will I also have to list of everything else that will affect the build?

Even so it's still unexpected that the things Cargo knows about the source layout isn't taken into account.

Canop:
It's additive.

> Even so it's still unexpected that the things Cargo knows about the source layout isn't taken into account.

Last time I checked, there was no good solution in the rust ecosystem for getting the information regarding the project's metadata. As I'd rather not maintain that part myself I use cargo-metadata whose design isn't great.

alerque:
Fair enough, and thanks for the info.

Before I opened this issue I had a quick look at the source to see if this was something I could easily fix, but since I didn't see if/where the manifest was being read at all it wasn't obvious how to fix this.

> Last time I checked, there was no good solution in the rust ecosystem for getting the information regarding the project's metadata.

[...]

PR #456, opened 2026-08-19
Watch build scripts relocated via the package.build manifest key

ChrisJr404:
This addresses #205. The default watch list hard-codes `build.rs`, so when a package moves its build script somewhere else with the `package.build` manifest key (for example `build = "build-aux/build.rs"`, as in the git-warp-time project mentioned in the issue), that file is never watched and editing it does not trigger a run.

Since bacon already loads the cargo metadata, I use it to find each local package's `custom-build` target and watch its real `src_path`. When the build script sits at the conventional location the path matches the existing `build.rs` watch and is deduplicated, so nothing changes for the common case. It is gated on `default_watch` too, so a job that opts out of the default watches is unaffected.

You mentioned wanting to look at the `build` part in cargo-metadata when you found the time, so hopefully this saves you some. I added a small unit test over the extraction and a CHANGELOG line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - fix crash in search on multibyte char in pattern and wrapped output - Fix #446
 - fix crash of search when output grows - Fix #447
 - fix unability to remove backtab binding - Fix #450
+- watch build scripts relocated with the `package.build` manifest key, instead of only the default `build.rs` path - Fix #205
 
 <a name="v3.24.0"></a>
 ### v3.24.0 - 2026/07/14
--- a/src/context.rs
+++ b/src/context.rs
@@ -182,6 +182,17 @@ impl Context {
                         } else {
                             warn!("missing manifest file: {:?}", item.manifest_path);
                         }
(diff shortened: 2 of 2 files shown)


Canop:
This looks good. Did you test it (I mean on a project with a `Cargo.toml` rather than just with the Target related unit tests) ?

ChrisJr404:
Yeah, I tested it against real projects, not just the unit tests. I built this branch and ran the resulting bacon binary (a `check` job) against three throwaway cargo projects, then read the actual watched paths back from bacon's own log with `BACON_LOG=info`.

The first project has `build = "build-aux/build.rs"` in its `[package]`. The watcher list came back with `.../build-aux/build.rs` in it, so editing the relocated script now triggers a run.

The second is a normal project with a conventional `build.rs`. It still shows up exactly once (the metadata path matches the default watch and gets deduplicated), so nothing changed for the common case.

The third has no build script at all, and its list is just `Cargo.toml` plus `src`, so it's unaffected.

So it behaves the way the PR describes on an actual Cargo.toml, not only in the Target unit tests.

PR #463, opened 2026-08-29
Remove some `flush`es when redrawing

narpfel:
`bacon` redraws the whole screen on every line of output received from the child process. When processing lots (as in multiple thousands of lines) of output, this takes a significant amount of time.

This PR removes some unnecessary `flush`es that occur during redraw.

On a (simple and unscientific) benchmark of printing 10_000 lines each containing the numbers from 0 to 100, this improves performance. The exact improvement depends on the terminal emulator, the window size (and probably more factors that I didn’t measure). E. g. in an 80×24 Alacritty window (the `empty` job is used to measure startup time):
```console
$ time ./bacon-main empty
./bacon-main empty  0,00s user 0,01s system 20% cpu 0,068 total
$ time ./bacon-main
./bacon-main  0,70s user 0,77s system 97% cpu 1,519 total
$ time ./bacon-reduce-flushing
./bacon-reduce-flushing  0,35s user 0,08s system 74% cpu 0,582 total
$ time ./produce-output
[... 10_000 lines]
./produce-output  0,01s user 0,04s system 50% cpu 0,087 total
```

Tested with this `bacon.toml`:
```toml
default_job = "produce-output"

[jobs.produce-output]
command = ["./produce-output"]
need_stdout = true
background = false
on_success = "quit"

[jobs.empty]
command = ["echo", "42"]
need_stdout = true
background = false
on_success = "quit"
```

### DIFF
--- a/src/tui/drawing.rs
+++ b/src/tui/drawing.rs
@@ -3,7 +3,7 @@ use {
     anyhow::Result,
     termimad::crossterm::{
         cursor,
-        execute,
+        queue,
         terminal,
     },
 };
@@ -14,7 +14,7 @@ pub fn goto(
     x: u16,
     y: u16,
 ) -> Result<()> {
-    execute!(w, cursor::MoveTo(x, y))?;
+    queue!(w, cursor::MoveTo(x, y))?;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


narpfel:
My use case for this amount of output is a parser written using [`lalrpop`](https://github.com/lalrpop/lalrpop), which outputs *a lot* of ... stuff when there are errors in the grammar. This makes `bacon` pretty much unusable when working on the grammar to e. g. debug ambiguities because it adds like 15 seconds to the edit-compile loop.

Canop:
Merged. This shouldn't really change the speed much for you.

I'll see if I can get bigger improvement on this kind of use cases (i.e. huge outputs). You may create an issue to better track it.

narpfel:
Thanks!

I poked around a bit with skipping `draw` calls when the additional output is offscreen, but being naive there breaks scrolling, so I kinda gave up. Will create an issue if/when I can’t figure something out after spending a bit more time on it.

Canop:
In fact I think I have a solution. Don't look for simple draws, it's more radical and it looks like it does greatly improve the situation on a sample program.

I'll try to finish it this week-end.

Canop:
Have a look at #464

It should improve speed a lot in your case.

(I still have to check it breaks nothing but I don't find flaws in the logic yet)

PR #464, opened 2026-09-04
drain all pending lines before a redraw

Canop:
This avoids drawing many times when lines come in burst


narpfel:
This looks great, thanks!

I’ll test this more thoroughly when I get the chance and report back if I notice anything.

PR #465, opened 2026-09-07
Go analyzer: fix test pass status

bitfield:
The standard analyzer sets the green 'pass!' job status based on the number of passed tests being > 0, but Go doesn't report passed tests, only failures. This commit modifies the Go analyzer to set the `has_test_passed` field to `true` if there are no errors or failures (Go doesn't have warnings).

The effect of this is that using `bacon` in Go projects now shows a green `pass!` flag for passing tests.


Canop:
> Go doesn't report passed tests, only failures

I don't think that's true. It's more than the current code of the analyzer doesn't read them. This should be fixed, which looks doable with a regex like `regex_is_match!(r"^ok\s+\S+\s+(?:\d+\.\d+s|\(cached\))", &raw)`.

bitfield:
Sounds good—would you like me to make that change?

Canop:
Yes, as I don't code in Go those days, I'm not really at ease doing this kind of change.

bitfield:
Thinking about this a bit more, it seems like there are really two related issues here:

1. Reporting the 'tests passed' status correctly (this PR)
2. Listing the passed tests in the same way that Rust does

Issue 1 is really a bug fix, because the Go analyzer _should_ be showing a green badge when all tests pass, but it isn't (this PR fixes that).

Issue 2 is really a new feature: something like “Go analyzer should list passing tests”. That is much more involved than fixing the test status bug, because the `go test` command only reports failures, not passes.

For example, here's the output of `go test` in a project of mine:

```
PASS
ok  	github.com/bitfield/go-machine	0.352s
```

And that's what Go programmers expect to see when they run tests: they generally don't want a list of all their passing tests. They just want one of two messages: (1) everything's fine, nothing to see here, (2) here is the list of failing tests.

We _can_ get that information by running `go test -v` and parsing all the output lines for success or failure in the same sort of way that the standard analyzer does. The question is whether you _want_ that behaviour, and whether it's helpful for Go programmers using `bacon`. I personally wouldn't use this feature, for example, but that's a data point, not an argument.

It might be good to merge _this_ fix first, and then open a new issue for the list-passing-tests feature so that people can weigh in with opinions about how it should work, or if they even want it. Let me know if this sounds like a good way to proceed.

Canop:
Not arguing, just trying to understand

> We can get that information by running go test -v and parsing all the output lines for success or failure in the same sort of way that the standard analyzer does. The question is whether you want that behaviour, and whether it's helpful for Go programmers using bacon. I personally wouldn't use this feature

What's the problem in having bacon call `go test -v` in order to know the number of tests ? There's no burden on the user. What's the downside ?

Canop:
At least check that the output is one of go test before setting has_passed_tests to true.

bitfield:
No downside that I'm aware of—that sounds like the right implementation for the feature you have in mind. I'd just prefer to keep this PR small, and focused on the minimal fix for the issue. I understand if you'd rather wait for the fuller solution that involves counting passed tests—if so, please feel free to close this.

Canop:
And what about the intermediate solution: not really counting test but checking that the output is of go test (ie with an output like the one you display in this thread) ?

bitfield:
That's a good point—we could look for "PASS"?

